PR #9, opened 2026-04-27
fix(netsync): harden sync-peer selection against forged version height

Aharonee:
## Summary

Sync-peer candidates are now restricted to outbound peers with `SFNodeNetwork`,
picked uniformly at random. The unauthenticated `peer.LastBlock()` from the
version handshake is no longer used for ranking — it's a forgeable 32-bit
integer that any connecting peer can set to an arbitrary value.

This mirrors Bitcoin Core's `CanServeBlocks` + `fPreferredDownload` posture
in `net_processing.cpp`'s `SendMessages`.

### Changes

- **`pickSyncCandidate`** replaces the height-ranked `higherPeers`/`equalPeers`
  partitioning in `startSync`. Filter: `syncCandidate` (services) + outbound +
  cooldown check. Uniform random pick from remaining pool.
- **Stall cooldown**: when `handleStallSample` rotates away from a stalled
  syncnode, the peer's address is excluded from re-selection for 10 minutes.
  Inbound peers are skipped (ephemeral source port).
- **Periodic retry**: `handleStallSample` retries `startSync` when no syncpeer
  exists and the chain is not current, so expired cooldown entries get
  re-evaluated without waiting for peer churn.
- Drive-by: fix failing `Example_blockStorageAndRetrieval` test (`db.Update` →
  `db.View` for read path, expected block size 533 → 612).
- Version bump 1.0.0 → 1.0.1.

### Test

`TestSyncPeerLiesAboutHeight` (`node/integration/sync_lie_test.go`, build tag
`rpctest`): inbound peer claims `LastBlock=999_999` with empty getblocks/getheaders
replies. Asserts it never appears as syncnode and the victim syncs from a
legitimate outbound peer within 30 s. Runtime ~3 s.

### Not in scope

- Headers-presync (post-selection PoW enforcement) — separate in-progress PR.
- Inv-triggered parallel `getheaders` (Bitcoin Core PR #25720) — belongs to
  the headers-presync PR.
- `addrmgr` flap-penalty for outbound-slot churn — separate workstream.
- Banscore for header-misbehavior disconnect paths — separate concern.

### DIFF
--- a/node/database/example_test.go
+++ b/node/database/example_test.go
@@ -129,7 +129,6 @@ func Example_blockStorageAndRetrieval() {
 	}
 	db, err := database.Create("ffldb", dbPath, wire.MainNet)
 	if err != nil {
-		fmt.Println("fail here")
 		fmt.Println(err)
 		return
 	}
@@ -151,7 +150,7 @@ func Example_blockStorageAndRetrieval() {
 	// Use the View function of the database to perform a managed read-only
 	// transaction and fetch the block stored above.
 	var loadedBlockBytes []byte
-	err = db.Update(func(tx database.Tx) error {
+	err = db.View(func(tx database.Tx) error {
 		genesisHash := chaincfg.MainNetParams.GenesisBlock.BlockHash()
 		blockBytes, err := tx.FetchBlock(&genesisHash)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


raskin1 (approved):
LGTM

idae577:
The wallet sucks. It generates a new address for receiving coins every time I access it. I haven't received any coins from mining yet. I just mined someone else's coins.

PR #10, opened 2026-04-27
updated workflow permissions and pinned actions to hash

raskin1:
## Summary

updated workflow permissions and pinned actions to hash.

Also refactored workflows into 3 tiers - Untrusted (always runs), Maintainer-approved and protected (runs only manually on master)

## Type

chore

## Scope

github

### DIFF
--- a/.github/actions/start-ec2-runner/action.yml
+++ b/.github/actions/start-ec2-runner/action.yml
@@ -42,7 +42,7 @@ runs:
   using: composite
   steps:
     - name: Configure AWS credentials
-      uses: aws-actions/configure-aws-credentials@v4
+      uses: aws-actions/configure-aws-credentials@7474bc4690e29a8392af63c5b98e7449536d5c3a # v4
       with:
         aws-region: ${{ inputs.aws-region }}
         aws-access-key-id: ${{ inputs.aws-access-key-id }}
@@ -51,7 +51,7 @@ runs:
     - name: Start EC2 runner
       id: start
       continue-on-error: true
-      uses: machulav/ec2-github-runner@v2
+      uses: machulav/ec2-github-runner@343a1b2ae682e681c3cec9a235d882da17ff04ef # v2
       with:
         mode: start
         github-token: ${{ inputs.github-token }}
--- a/.github/workflows/blockchain_ci.yml
+++ b/.github/workflows/blockchain_ci.yml
@@ -1,8 +1,6 @@
 name: Blockchain / Build and Test
 on:
   pull_request:
-    branches:
-      - master
     paths:
       - 'node/**'
       - 'wallet/**'
@@ -17,6 +15,10 @@ on:
 
 permissions:
   contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
 
 env:
   # go needs absolute directories, using the $HOME variable doesn't work here.
@@ -74,3 +76,14 @@ jobs:
         run: task test:go GOTESTSUM_FORMAT=pkgname-and-test-fails
         env:
           GOPRIVATE: github.com/pearl-research-labs/*
+
+  ci_status:
+    name: Blockchain CI Status
+    runs-on: ubuntu-latest
+    needs: [format-tidy, build-and-test]
+    if: always()
+    steps:
+      - run: |
+          result="${{ (contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled')) && 'failure' || 'success' }}"
+          echo "Overall result: $result"
+          [ "$result" = 'success' ]
--- a/.github/workflows/blockchain_ci.yml
+++ b/.github/workflows/blockchain_ci.yml
@@ -69,13 +69,9 @@ jobs:
 
       - name: Build
         run: task build:blockchain
-        env:
-          GOPRIVATE: github.com/pearl-research-labs/*
 
       - name: Test
         run: task test:go GOTESTSUM_FORMAT=pkgname-and-test-fails
-        env:
-          GOPRIVATE: github.com/pearl-research-labs/*
 
   ci_status:
     name: Blockchain CI Status
--- a/.github/workflows/blockchain_ci.yml
+++ b/.github/workflows/blockchain_ci.yml
@@ -15,6 +15,8 @@ on:
       - '.github/actions/setup/action.yml'
   workflow_dispatch:
 
+permissions:
+  contents: read
 
 env:
   # go needs absolute directories, using the $HOME variable doesn't work here.
@@ -27,7 +29,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          persist-credentials: false
 
       - name: Setup Environment
         uses: ./.github/actions/setup
@@ -37,7 +41,7 @@ jobs:
 
       - name: Check for Changed Files
         id: verify-changed
-        uses: tj-actions/verify-changed-files@v20.0.4
+        uses: tj-actions/verify-changed-files@a1c6acee9df209257a246f2cc6ae8cb6581c1edf # v20.0.4
 
       - name: Fail if files changed
         if: steps.verify-changed.outputs.files_changed == 'true'
@@ -54,7 +58,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          persist-credentials: false
 
       - name: Setup Environment
         uses: ./.github/actions/setup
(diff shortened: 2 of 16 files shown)


o-k-d (dismissed):

.github/actions/start-ec2-runner/action.yml
```diff
@@ -42,7 +42,7 @@ runs:
   using: composite
   steps:
     - name: Configure AWS credentials
-      uses: aws-actions/configure-aws-credentials@v4
+      uses: aws-actions/configure-aws-credentials@7474bc4690e29a8392af63c5b98e7449536d5c3a # v4
```
matan-dup, line 45: Are these changes actually meaningful?

.github/workflows/blockchain_ci.yml
```diff
@@ -27,7 +31,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
```
matan-dup, line 34: I really think that `actions/checkout@v6` is no less secure.

.github/actions/start-ec2-runner/action.yml
raskin1: a supply chain attack can change the v4 pointer to something else. seems this is the best practice currently.
dependabot knows how to work with this format and keep it updated

liad-madmon (dismissed):

.github/workflows/blockchain_ci.yml
```diff
@@ -68,3 +76,14 @@ jobs:
         run: task test:go GOTESTSUM_FORMAT=pkgname-and-test-fails
         env:
           GOPRIVATE: github.com/pearl-research-labs/*
```
Aharonee: No need for GOPRIVATE directives any more since the repo is public

.github/workflows/miner_gpu_ci.yml
```diff
@@ -12,22 +13,36 @@ permissions:
...
 
 env:
   UV_PROJECT_ENVIRONMENT: ${{ github.workspace }}/.venv-gpu-ci
+  PR_HEAD_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
 
 jobs:
+  gate:
+    name: Check Approval Gate
+    runs-on: ubuntu-latest
+    if: >
```
Aharonee: The job-level `if:` makes this gate **skip** rather than **fail** when criteria aren't met. A skipped job propagates as `skipped` (not `failure`) through `needs:`, which silently passes any future downstream job that uses `if: always()` and doesn't short-circuit on an output check. The current pipelines work only by accident of empty-string output checks (`needs.detect-changes.outputs.should_run == 'true'`).

Suggest making the gate always run and explicitly fail:

```yaml
gate:
  runs-on: ubuntu-latest
  steps:
    - name: Verify approval
      run: |
        if [[ "${{ github.event_name }}" == "merge_group" ]]; then exit 0; fi
        if [[ "${{ github.event_name }}" == "pull_request_target" ]] \
           && [[ "${{ contains(github.event.pull_request.labels.*.name, 'safe-to-test') }}" == "true" ]]; then exit 0; fi
        echo "::error::Tier 2 requires merge_group or 'safe-to-test' label"
        exit 1
```

Same pattern applies in `integration_tests_ci.yml`, `miner_heavy_ci.yml`, and `pearl-desktop-wallet.yml`.

.github/workflows/miner_ci.yml
```diff
@@ -1,17 +1,12 @@
-name: miner CI
+name: "Miner CI: Lint & Unit Tests"
 
 on:
   pull_request:
```
Aharonee, line 4: **This breaks the merge queue.** `master`'s ruleset requires `Miner CI Status` and `Rust CI Status`. This PR removed the `merge_group` trigger from both workflows (this one + `rust_ci.yml`), so those checks never report on the merge-group commit and the queue will hang for 120 min before failing every merge attempt.

Fix: re-add `merge_group: { types: [checks_requested] }` to both files.

.github/workflows/pearl-desktop-wallet.yml
```diff
@@ -143,9 +156,11 @@ jobs:
...
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          persist-credentials: false
+          ref: ${{ env.PR_HEAD_SHA }}
 
-      # ========== Build Go Binaries ==========
       - name: Setup Go and Rust
```
Aharonee, line 166: CodeQL #30 (cache poisoning) — `pull_request_target` cache writes go to **master's cache scope** (not branch-isolated like `pull_request`), so a fork PR running this composite can populate caches that master and the wallet release pipeline later read from.

The composite enables two caches; they have different threat profiles:

**`actions/setup-go` (`cache: true`)** — Go's build cache is content-addressed by source+flags+deps, so master can't restore objects whose source hash doesn't match. The `cache: true` is fine to keep; **dismiss CodeQL alert #30** with that rationale.

**`Swatinem/rust-cache@v2`** — Rust's incremental cache is *not* content-addressed in the same way. A malicious PR can populate poisoned `.rlib` files in `target/` that subsequent master/release Rust builds happily link in. Salt the cache key with `event_name` so PR caches don't share scope with master:

```yaml
# in .github/actions/setup/action.yml
- name: Set up Rust cache
  uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2
  with:
    prefix-key: ${{ github.event_name }}
    workspaces: |
      zk-pow -> target
      zk-pow/bindings/go -> target
    cache-on-failure: true
```

This preserves cache-hit performance within each scope (master shares with master, PR runs share with PR runs) while preventing cross-scope poisoning. Same composite is invoked from `integration_tests_ci.yml`, `miner_gpu_ci.yml`, `miner_heavy_ci.yml`, and `release.yml` — fix is one place.

Aharonee (changes requested):
**Three changes needed in `.github/actions/setup/action.yml`** (the composite invoked from `blockchain_ci.yml`, `release.yml`, and all Tier-2 workflows). The file isn't in this PR's diff, so leaving this here as a top-level review note rather than inline.

**1. SHA-pin all four actions** — consistency with the rest of this PR; also clears the Node 20 deprecation warning currently raised against `setup-go@v5` and `setup-task@v2` in CI. SHAs already in use elsewhere in this PR:
- `actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5` (already used in `release.yml`)
- `dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8 # stable`
- `Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2`
- `arduino/setup-task@<pin>` — needs a fresh SHA, not yet used elsewhere in this PR.

**2. Remove the `Configure Git for GitHub module fetching` step.** The `git config --global url...insteadOf` was only needed for fetching private `pearl-research-labs/*` modules. None of the three `go.mod` files require any such modules and the repo is public. Removing it also restores the integrity of `persist-credentials: false` across Tier-2 workflows (currently undone the moment this composite runs and writes `${{ github.token }}` into `~/.gitconfig`).

**3. Salt `Swatinem/rust-cache` with `prefix-key: ${{ github.event_name }}`** — closes the cache-poisoning vector noted on `pearl-desktop-wallet.yml:164`. One-line change here covers all callers (`pearl-desktop-wallet.yml`, `integration_tests_ci.yml`, `miner_gpu_ci.yml`, `miner_heavy_ci.yml`, `release.yml`).

.github/workflows/blockchain_ci.yml
raskin1: removed

.github/workflows/miner_gpu_ci.yml
raskin1: done

.github/workflows/pearl-desktop-wallet.yml
raskin1: fixed

.github/workflows/miner_ci.yml
raskin1: fixed

Aharonee (dismissed):

Aharonee:
Follow-up, separate from the earlier review: `id-token: write` in `integration_tests_ci.yml` and `miner_gpu_ci.yml` is dead permission and worth removing.

Verified — no consumer of OIDC tokens anywhere in `.github/`: no `role-to-assume`, no `getIDToken`, no `workload_identity`, no `cosign`/`sigstore`. Both files' AWS auth uses static `aws-access-key-id` + `aws-secret-access-key`, not OIDC. The permission pre-dates this PR (added in the initial Pearl commits, likely cargo-culted from a reference workflow) and removing it doesn't change any current behavior.

Under `pull_request_target`, PR-controlled code can call `core.getIDToken()` to mint OIDC JWTs. If any AWS/GCP/Azure trust policy is later configured to accept this workflow's tokens (without realizing it runs PR-controlled code), it becomes exploitable. Removing the permission closes that future-footgun and tightens the residual concern flagged by CodeQL #26.

```yaml
# .github/workflows/miner_gpu_ci.yml
# .github/workflows/integration_tests_ci.yml
permissions:
-  id-token: write
   contents: read
   pull-requests: read
   packages: read
```

After removal, CodeQL #26 can be dismissed with rationale: *label-gated by `safe-to-test`, ephemeral runner, no mintable federation token, no secrets referenced in the PR-head step.*

Aharonee (approved):

PR #29, opened 2026-04-28
fix(netsync): fall back to inbound peers for sync during IBD

Aharonee:
## Summary

When no outbound sync-peer candidate is available and the chain is not
current, `pickSyncCandidate` now falls back to inbound peers. The
`handleInvMsg` gate is also relaxed to accept block announcements from
any peer when `syncPeer` is nil.

Without this, a node whose only block source connects inbound (e.g. a
NAT node dialling anchors) deadlocks at genesis: `pickSyncCandidate`
returns nil (no outbound), `handleInvMsg` drops the inbound peer's
block invs, and chain progress stays at zero.

Mirrors Bitcoin Core's `sync_blocks_and_headers_from_peer` fallback
(PR [#24171](https://github.com/bitcoin/bitcoin/pull/24171), merged
June 2022). Outbound peers remain preferred when available.

### Changes

- `pickSyncCandidate`: partition candidates into outbound/inbound
  buckets; use inbound only when outbound is empty and `!IsCurrent()`.
- `handleInvMsg`: allow invs through when `syncPeer == nil`, so blocks
  from inbound peers can be discovered during IBD.
- `TestSyncPeerLiesAboutHeight`: connect honest (outbound) before liar
  (inbound) so the test correctly asserts inbound exclusion when an
  outbound candidate exists.
- New `TestSyncFromInboundDuringIBD`: source dials victim inbound,
  mines 10 blocks, victim syncs within 30 s. Verified: fails without
  the fix (30 s timeout), passes with it (1 s).
- Version bump 1.0.1 → 1.0.2.

### Test

```
go test -tags=rpctest -timeout 5m -run 'TestSyncFromInboundDuringIBD|TestSyncPeerLiesAboutHeight|TestSyncManagerRaceCorruption|TestPreVerackDisconnect' ./node/integration
```

### DIFF
--- a/node/netsync/manager.go
+++ b/node/netsync/manager.go
@@ -264,22 +264,23 @@ func (sm *SyncManager) findNextHeaderCheckpoint(height int32) *chaincfg.Checkpoi
 // pickSyncCandidate returns a random sync-peer candidate, filtered by:
 //   - state.syncCandidate (SFNodeNetwork / pruned-with-block-threshold,
 //     set at handshake by isSyncCandidate).
-//   - Outbound only. peer.LastBlock() from the version handshake is
+//   - Outbound preferred. peer.LastBlock() from the version handshake is
 //     unauthenticated; restricting to outbound peers limits candidates
 //     to addresses we chose to dial (addr.dat / dnsseed).
 //   - recentlyFailedSync cooldown to prevent immediate re-selection of
 //     peers that previously stalled as syncnode.
 //
-// peer.LastBlock() is intentionally not used for ranking. Mirrors Bitcoin
-// Core's CanServeBlocks + fPreferredDownload posture (net_processing.cpp).
+// When no outbound candidate is available and the chain is not current,
+// inbound peers are accepted as a fallback so that a node whose only
+// block source is inbound (e.g. a NAT node dialling us) can still sync
+// during IBD.
+//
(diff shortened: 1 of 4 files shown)


node/netsync/manager.go
```diff
@@ -1153,9 +1164,10 @@ func (sm *SyncManager) handleInvMsg(imsg *invMsg) {
 		peer.UpdateLastAnnouncedBlock(&invVects[lastBlock].Hash)
 	}
 
-	// Ignore invs from peers that aren't the sync if we are not current.
-	// Helps prevent fetching a mass of orphans.
-	if peer != sm.syncPeer && !sm.current() {
+	// Ignore invs from non-syncpeers when not current to prevent
+	// fetching orphans. Allow through when syncPeer is nil so blocks
+	// reachable only via inbound peers can be discovered during IBD.
+	if peer != sm.syncPeer && !sm.current() && sm.syncPeer != nil {
```
o-k-d: Pointing out that [not current and syncPeer == nil] is a strange scenario and handling invs will not help. No action item though...

o-k-d (dismissed):

node/netsync/manager.go
Aharonee: There's a potential short window where this condition could be satisfied, however on second thought I can't see a strong justification for this change. Reverted.

o-k-d (approved):

PR #66, opened 2026-05-04
fix(wallet/walletdb): assign error return from walletdb.View

ygd58:
## Summary

Fixes errcheck lint violation found by `task lint:go`.

`walletdb.View()` return value was discarded in `testReadWriteInterface`. The `err` variable was declared earlier in the function but not assigned here, meaning any database error during top-level bucket enumeration would be silently ignored.

## Testing
- `go build ./wallet/...` ✅ passes


ygd58:
Superseded by aggregated PR as requested by @Aharonee

PR #84, opened 2026-05-06
fix(connmgr): preserve retry backoff across short-lived connections

Aharonee:
## Summary

`maxRetryDuration` was unreachable in practice for permanent peers. `retryCount` was reset to zero inside `handleConnected` on every TCP `Dial` success, before the application layer could decide whether to keep the connection. A remote that accepts TCP and then immediately drops the connection (for any post-handshake reason) caused the cycle to pin at `retryCount=1`, so the linear multiplier never grew and the documented 5-minute cap never engaged.

This change ties the reset to actual connection lifetime instead of bare TCP success.

## Changes in `node/connmgr/connmanager.go`

- New package var `stableConnectionDuration` (default 30s): minimum lifetime before a connection is considered a real success.
- New `ConnReq.establishedAt` field, stamped in `handleConnected` and read in `handleDisconnected`.
- `handleConnected` no longer resets `retryCount`; it records `establishedAt` instead.
- `handleDisconnected` resets `retryCount` only if `time.Since(establishedAt) >= stableConnectionDuration`. Briefer drops keep the counter intact so backoff continues to grow toward `maxRetryDuration`.
- All `retryCount` mutations switched to atomic operations so external readers (tests, future observability hooks) can load it safely from outside the `connHandler` goroutine.

## Tests in `node/connmgr/connmanager_test.go`

Three new in-package tests:

- `TestRetryCountGrowsOnInstantDrops` — regression test. Mocked dialer returns a real conn; `OnConnection` immediately calls `Disconnect`, simulating a remote that accepts TCP and drops at the application layer. Asserts `retryCount` grows past 1 across cycles. Fails on master without this fix.
- `TestMaxRetryDurationCapEngages` — verifies the existing `d > maxRetryDuration` clamp actually engages once `retryCount` exceeds the cap engagement threshold.
- `TestRetryCountResetsAfterStableConnection` — positive case. Pre-loads `retryCount=7`, holds the connection past `stableConnectionDuration`, then disconnects; asserts the reset fires.

`init()` also sets `stableConnectionDuration=10ms` for the test binary; mutating these vars from inside a test races with the `connHandler` goroutine, so they are pinned at startup.

## Verification

- `go vet ./node/connmgr/...`: clean
- `go test ./node/connmgr/`: 18/18 (3 new + 15 existing)
- `go test -race -count=10 ./node/connmgr/`: clean across all 10 iterations

## Test plan

- [x] Run `go test -race ./node/connmgr/` and confirm all tests pass.
- [x] Toggle off the new gated reset in `handleDisconnected` and confirm `TestRetryCountGrowsOnInstantDrops` fails as expected, demonstrating the regression coverage.

### DIFF
--- a/node/connmgr/connmanager.go
+++ b/node/connmgr/connmanager.go
@@ -32,6 +32,15 @@ var (
 	// persistent connections.
 	defaultRetryDuration = time.Second * 5
 
+	// stableConnectionDuration is the minimum lifetime a connection must
+	// reach before its eventual disconnect resets the permanent-peer
+	// retry counter. TCP connect succeeds at the kernel layer well before
+	// the application can decide whether to keep the peer; without this
+	// floor a remote that closes immediately after accept (e.g. it is at
+	// peer capacity) would silently zero the backoff and pin retries at
+	// the base interval forever.
+	stableConnectionDuration = time.Second * 30
+
 	// defaultTargetOutbound is the default number of outbound connections to
 	// maintain.
 	defaultTargetOutbound = uint32(8)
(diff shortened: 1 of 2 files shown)


o-k-d (approved):
💪

PR #87, opened 2026-05-08
fix(miner/pearl-gateway): fix type annotation errors

ygd58:
Fixes type annotation errors found by mypy in pearl-gateway:

- `pearl_client.py`: `params` default `None` typed as `list[Any] | None`; `session` field annotated as `aiohttp.ClientSession | None`
- `scheduler.py`: `last_refresh_time` initialized as `float` to match `time.time()` return type
- `submission_service.py`: `submission_log` typed as `set[bytes]` since `serialize_without_proof_commitment()` returns `bytes`


ygd58:
Closing — these fixes were AI-assisted rather than from real usage experience. Will focus on contributions that come from actually running and understanding the system.

PR #88, opened 2026-05-08
fix(miner/miner-base): fix type errors in async_loop_manager

ygd58:
Fixes two type errors in async_loop_manager.py found by mypy:

- Added `assert self._loop is not None` before `call_soon_threadsafe` call to guard against uninitialized event loop
- Added `self._mining_job is not None` guard before accessing `incomplete_header_bytes` and `target` in logging branch


ygd58:
Closing — same reason as #87.

PR #89, opened 2026-05-08
feat(rpc): add ProofCommitment to getblock verbose response

ygd58:
Pearl blocks contain a ZK proof commitment in the block header, but `GetBlockVerboseResult` and `GetBlockVerboseTxResult` did not expose this field in the `getblock` RPC response.

This PR adds `proofcommitment` to the verbose block output so RPC clients and block explorers can access it directly.

**Before:**
```json
{ "hash": "...", "height": 9728, ... }
```

**After:**
```json
{ "hash": "...", "height": 9728, "proofcommitment": "21e77d33...", ... }
```

Found by running a testnet node and querying `getblock`.

### DIFF
--- a/node/btcjson/chainsvrresults.go
+++ b/node/btcjson/chainsvrresults.go
@@ -28,8 +28,9 @@ type GetBlockHeaderVerboseResult struct {
 	Time          int64   `json:"time"`
 	Bits          string  `json:"bits"`
 	Difficulty    float64 `json:"difficulty"`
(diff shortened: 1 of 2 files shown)


Aharonee (dismissed):
Looks good after formatting and adding the missing RPC help descriptions for the new proofcommitment fields.

Verified locally:
- task fmt tidy
- go test ./node/btcjson ./node

PR #91, opened 2026-05-10
feat(netsync): add per-peer quality gate for inv-driven block download

o-k-d:
## Summary
Adds a per-peer quality gate in `netsync` so low-quality peers cannot turn block `inv` announcements directly into full-block downloads.

High-quality peers keep the existing fast path. Low-quality peers are asked for cert-less headers first; only headers that connect to a block we already know are converted into targeted `getdata` requests.

## Before
Every sync-eligible peer's first block `inv` could trigger `getdata`, so a fabricated orphan cost the peer one inv vector and cost us a full block download plus orphan handling:

```
inv(hash) -> getdata -> full block
                    -> ProcessBlock returns isOrphan = true
                    -> addOrphanBlock (bounded; no peer quality signal)
                    -> PushGetBlocksMsg(orphanRoot)
```

The orphan pool cap bounds memory, but without a per-peer quality signal the same peer could keep spending our bandwidth and validation work over the lifetime of the connection.

## What changed
1. **Per-peer quality signal.** `peerSyncState.nonTipStrikes` tracks consecutive non-tip outcomes. Inbound peers start low-quality; outbound peers start high-quality. A block that lands on the main chain resets strikes to `0`; orphan, side-chain, and validation-error outcomes move the peer toward low-quality. Duplicate blocks are not penalized.

2. **Gated inv handling.** When the node is current, a low-quality peer's block announcements are routed through one cert-less `getheaders` probe per inv batch, anchored on the last announced block. If we already know that anchor, no probe is sent. Block invs from low-quality peers then skip the per-inv block path, so they cannot drive orphan retries or long-side-chain stall handling directly.

3. **Cert-less headers response handling.** Outside headers-first mode, `handleHeadersMsg` accepts only cert-less headers, requires the first header's `PrevBlock` to be a block we know, and then performs a single pass over the batch to request only headers we do not already have and have not already requested.

4. **Headers-first hardening.** Headers-first sync ignores non-sync-peer headers and validates certificate-bearing headers before mutating header state.

## Why this is sufficient
- A high-quality peer can cause at most `lowQualityStrikeLimit` consecutive non-tip block downloads before it is downgraded. Regaining quality requires delivering a block that lands on our main chain.
- A low-quality peer's orphan announcements no longer reach immediate `getdata`; they must first produce a headers response rooted in a block we know.
- If a peer does convince us to request a block, invalid or non-tip results feed back into the same strike logic, and existing block validation / ban-score behavior still applies.

## Validation
- `go test ./netsync/... ./peer/... -count=1`
- `go build -tags rpctest ./integration/...`


node/netsync/manager.go
```diff
@@ -819,6 +852,14 @@ func (sm *SyncManager) handleBlockMsg(bmsg *blockMsg) error {
 		heightUpdate = best.Height
 		blkHashUpdate = &best.Hash
 
+		// Only a tip-extending block resets strikes; accepted-but-
+		// not-tip blocks drift the peer toward low quality.
+		if best.Hash == *blockHash {
+			state.nonTipStrikes = 0
+		} else {
+			state.strikeNonTip()
+		}
```
Aharonee: Consider deriving strike accounting from `ProcessBlock`'s own `isMainChain` return rather than three separate `state.strikeNonTip()` sites (lines 792, 830, 860) plus this re-derived `best.Hash == *blockHash` check.

The first return value of `ProcessBlock` is discarded at line 774, and the question it answers — did this block become part of the main chain? — is reconstructed here from chain state. That forces a manual `strikeNonTip()` call in each non-tip branch (error / orphan / accepted-but-not-tip).

Reading the value `ProcessBlock` already returns collapses it to one decision at the top:

```go
isMainChain, isOrphan, err := sm.chain.ProcessBlock(bmsg.block, behaviorFlags)
if isMainChain {
    state.nonTipStrikes = 0
} else {
    state.strikeNonTip()
}

if err != nil {
    // existing logging + reject path
    return err
}
// ... orphan handling, height updates, etc.
```

Why it's worth doing:

1. **Single source of truth.** `connectBestChain` already computes "this block extended the main chain" (`accept.go:73`); the discarded return is exactly that signal.
2. **Slight correctness gain.** Edge case: this block enters the main chain, then `processOrphans` promotes an orphan above it before `ProcessBlock` returns. `isMainChain == true` but `best.Hash != *blockHash`, so today we strike a peer whose block did contribute. The proposed version credits them correctly.
3. **Future-proof.** New failure branches need to remember to call `strikeNonTip()` today; with strike-by-default-unless-`isMainChain`, they're correct by construction.

Minor: DB-corruption `panic(dbErr)` (line 789) currently runs before the strike call; under the proposed ordering the strike lands first. Harmless (we panic immediately after) but worth a glance.

node/netsync/manager.go
```diff
@@ -943,31 +984,69 @@ func (sm *SyncManager) fetchHeaderBlocks() {
...
-		log.Warnf("Got %d unrequested headers from %s -- "+
-			"disconnecting", numHeaders, peer.Addr())
-		peer.Disconnect()
-		return
-	}
 
 	// Nothing to do for an empty headers message.
 	if numHeaders == 0 {
 		return
 	}
```
Aharonee: Hoisting the empty-headers no-op to the top of the function, together with the three other silent-return paths added in this PR (`!havePrev` at L1012, link-break at L1024, `peer != sm.syncPeer` at L1046), removes a consequence the pre-PR code had: any headers message in non-headers-first mode — including empty — triggered a disconnect.

Worth flagging as a deliberate behavioral change rather than letting it land implicitly. Probably fine in steady state — the strike system catches the actually-expensive misbehavior via `ProcessBlock` failures, and the cheap cases (empty headers, junk that doesn't chain) are bandwidth-symmetric with no amplification. But a brief comment here confirming "intentional, cheap on both sides, no consequence applied" or a generic per-peer rate limit on headers messages would make the choice explicit.

Not advocating for peer-state tracking to strike on empty responses — distinguishing "lying about an inv" from "honestly caught up" is racy (our tip can advance between probe send and peer response, producing a legitimately-empty response from an honest peer).

node/netsync/manager.go
```diff
@@ -943,31 +984,69 @@ func (sm *SyncManager) fetchHeaderBlocks() {
...
+		root := msg.Headers[0].BlockHeader.PrevBlock
+		havePrev, errPrev := sm.chain.HaveBlock(&root)
+		if errPrev != nil || !havePrev {
+			return
+		}
+		firstHash := msg.Headers[0].BlockHeader.BlockHash()
+		haveFirst, errFirst := sm.chain.HaveBlock(&firstHash)
+		if errFirst != nil || haveFirst {
+			return
+		}
```
Aharonee: This early-out avoids striking a peer with `ErrDuplicateBlock` when they re-announce a block we already have — the request loop below dedupes against in-flight `requestedBlocks` (line 1032) but not against the chain. Pre-`3d1f09e1`, `requestBlocksFromHeaders` did per-header `HaveBlock` filtering inside the loop, which handled the same case at finer granularity.

The current shape aborts the entire batch on any "we have the first header." Corner case it gets wrong: peer responds with `[known, new_1, new_2]` — `havePrev` passes, `haveFirst` fires, the new blocks are silently dropped along with the duplicate. Rare on the current probe path (a getheaders response shouldn't include the locator's tip), but more plausible for the future BIP130 announce use the comment at L1005–1008 anticipates.

Restoring the per-header check inside the request loop fixes both:

```go
for i := range msg.Headers {
    hash := msg.Headers[i].BlockHeader.BlockHash()
    if _, exists := sm.requestedBlocks[hash]; exists {
        continue
    }
    if have, err := sm.chain.HaveBlock(&hash); err != nil || have {
        continue
    }
    ...
}
```

Same protection against unfair `ErrDuplicateBlock` strikes, no information loss on mixed batches, and `havePrev` is sufficient as the "must connect to our chain" anchor.

Aharonee (commented):
Some edge cases and unintended behavioral changes we should consider, but overall looks good (some code duplication but we'll handle it in the presync or a later PR)

node/netsync/manager.go
o-k-d: Done

node/netsync/manager.go
o-k-d: Done by going backward, with minimal block index locks.

node/netsync/manager.go
o-k-d: Commented that it is deliberate

Aharonee:
**Refactor suggestion: hoist the low-quality probe out of `handleInvMsg`'s loop**

The cert-less probe at L1332–1348 currently fires from inside the per-inv loop, guarded by a `sentGetHeaders` flag that exists only to enforce "at most one probe per inv batch." The stop hash is `&zeroHash`, so the probe isn't actually tied to any specific announced block — it asks the peer for everything above our locator.

Lifting it out and anchoring on `lastBlock` removes the flag, tightens the stop hash, and makes the gate uniform across block invs:

```go
if lastBlock >= 0 && sm.current() && !isPeerHighQuality(state) {
    if have, _ := sm.haveInventory(&invVects[lastBlock]); !have {
        locator, err := sm.chain.LatestBlockLocator()
        if err == nil {
            _ = peer.PushGetHeadersMsg(
                locator, &invVects[lastBlock].Hash, false)
        }
    }
}

for i, iv := range invVects {
    // existing type filter + AddKnownInventory ...
    if sm.headersFirstMode {
        continue
    }

    // Low-quality + current: the probe above covers all block invs
    // in this batch; skip them here uniformly.
    if (iv.Type == wire.InvTypeBlock ||
        iv.Type == wire.InvTypeWitnessBlock) &&
        sm.current() && !isPeerHighQuality(state) {
        continue
    }

    // existing tx / haveInv / orphan handling ...
}
```

Wins:

1. **One dispatch decision.** `sentGetHeaders` goes away — "at most one probe per batch" becomes structural rather than enforced by mutable state in the loop.
2. **Tighter stop hash.** Bounds the reply to the chain segment the peer actually claimed, not "everything above our locator up to 2000 headers." Still a hint (peer can ignore it), but the right hint.
3. **No probe when we already have `lastBlock`** — preserves today's `if !haveInv` short-circuit semantics.
4. **Threat-model consistency.** Today's gate only triggers on *unknown* block invs from low-quality peers; the known-block path still drives `IsKnownOrphan` orphan-root retrieval and the long-side-chain stall detection. With this shape, all block invs from low-quality-while-current peers are uniformly gated and those announcements don't get to drive our state machine.

Preservation notes for whoever implements:

- `peer.AddKnownInventory(iv)` must still run for skipped block invs — put the skip *after* it, not before, otherwise dedupe breaks on re-announces.
- `peer.UpdateLastAnnouncedBlock` / `UpdateLastBlockHeight` above the loop already cover their use cases.
- The `IsKnownOrphan` and stall-detection paths in the `if haveInv` branch become unreachable for low-quality peers; worth a one-line code comment confirming that's intentional per the threat model.

One honest-peer edge case the simple form doesn't cover: an out-of-order announce `[unknown, ..., known]` skips the probe and we don't learn about the earlier unknown from this peer this round. Real blocks will reach us via another peer or a re-announce, so it's deferred-learning, not missed-data — fine to accept.

Bonus: the existing gate at L1335 checks only `InvTypeBlock`. `InvTypeWitnessBlock` is also a valid wire type that peers could in principle use in inv messages; the refactor folds both naturally.

PR #100, opened 2026-05-12
Add presync mechanism

o-k-d:
Presync: headers validation for low-work chains

Without presync, a peer announcing a long chain of headers below the node's anti-DoS work threshold can trick the node into downloading full blocks (with certificates) for that chain -- an expensive operation in both bandwidth and disk I/O. Since Pearl blocks carry ZK proof certificates, downloading a block is significantly more costly than receiving a cert-less header.

Presync is a two-phase per-peer protocol that validates a claimed chain cheaply before committing to full block downloads:

Phase 1 (PRESYNC): The node receives cert-less headers from the peer, performing lightweight contextual validation (difficulty, timestamps) while storing only 1-bit SipHash commitments per header. Periodically, the node requests random spot-check headers with full certificates and verifies them. This bounds the probability of an attacker faking a long chain to `2^-100`, using minimal memory (~1 bit per header).

Phase 2 (REDOWNLOAD): Once the claimed chain accumulates sufficient work and all spot-checks pass, the node re-requests headers with full certificates and downloads blocks. During redownload, each header is verified against the commitment bits stored in phase 1, ensuring the peer cannot switch to a different chain mid-download. Blocks are processed with BFNoAntiDoSWork to avoid recursive presync initiation.

During IBD, headersFirstMode restricts presync to the single sync peer. After IBD, presync can run concurrently with multiple peers announcing low-work chains.

TODO: consider disconnecting from a peer whose presync failed noncritically -- although it might be just an honest misunderstanding.


node/chaincfg/params.go
```diff
@@ -380,6 +385,8 @@ var MainNetParams = Params{
 	// BIP44 coin type used in the hierarchical deterministic path for
 	// address generation.
 	HDCoinType: HDCoinTypePearl,
+
+	MinimumChainWork: new(big.Int).Lsh(big.NewInt(1), 85),
```
jonathanMweiss: Please add a comment stating where 2^85 came from.

I guess it should be around 2^88 now?

node/blockchain/accept.go
```diff
@@ -34,6 +35,20 @@ func (b *BlockChain) maybeAcceptBlock(block *btcutil.Block, flags BehaviorFlags)
 		return false, ruleError(ErrInvalidAncestorBlock, str)
 	}
 
+	// Anti-DoS: refuse to store a block whose candidate chain has not yet
+	// proven enough cumulative work. BFNoAntiDoSWork is set during
+	// presync REDOWNLOAD where anti-DoS is achieved through commitment bits.
+	if flags&BFNoAntiDoSWork != BFNoAntiDoSWork {
```
jonathanMweiss: Nit, non-blocking: this reads as a double negative ("flag is not not-set"). Something like `enforceAntiDoS := flags&BFNoAntiDoSWork == 0` before the `if` would make it easier to read.

jonathanMweiss: nit: `node/cmd/addblock/import:132` needs to pass flags to disabling antiDos. without it, datatdirs on main-net that have smaller than 2^85 cannot use it.

node/chaincfg/params.go
jonathanMweiss: Nit: this also has an issue with the addblock binary. some specific cases break there because it needs to receive flags to disable antiDos. 
The other option is to delete this specific binary.

node/blockchain/chain.go
```diff
@@ -2302,6 +2309,64 @@ func New(config *Config) (*BlockChain, error) {
...
+	}
+	return &ChainStartInfo{
+		Hash:          node.hash,
+		Height:        node.height,
+		Bits:          node.bits,
+		Timestamp:     node.timestamp,
+		WorkSum:       new(big.Int).Set(node.workSum),
+		PrevTimestamp: prevTs,
+	}
+}
```
jonathanMweiss: `ChainStartInfo` seems like a redundant struct and copying.

can return the blockNode directly as an interface: 
```Go
type ChainNode interface {
	chaincfg.HeaderCtx 
	WorkSum() *big.Int
}

func (node *blockNode) WorkSum() *big.Int {
	return new(big.Int).Set(node.workSum)
}

func (b *BlockChain) LookupChainNode(hash *chainhash.Hash) ChainNode {
	b.chainLock.RLock()
	defer b.chainLock.RUnlock()

	if 	node := b.index.LookupNode(hash); node != nil {
        // unfortunately can't just do return b.index.LookupNode(hash).
		return node
	}

	return nil
}
```

node/blockchain/difficulty.go
```diff
@@ -222,3 +222,69 @@ func (b *BlockChain) CalcNextRequiredDifficulty(timestamp time.Time) (uint32, er
 	b.chainLock.Unlock()
 	return difficulty, err
 }
+
+// paramsChainCtx is a minimal ChainCtx backed only by *chaincfg.Params, used
+// by callers that maintain lightweight state and do not have a full
+// *BlockChain (e.g. CalcNextRequiredDifficultyFromValues).
+type paramsChainCtx struct{ p *chaincfg.Params }
```
jonathanMweiss: This struct has odd behavior. and feels like a mock rather than actual logic. 
the core user of this struct is HeadersSyncState which has on creation time access to the chain itself.

Not sure what the benefit of adding this struct specific.

node/blockchain/difficulty.go
```diff
@@ -222,3 +222,69 @@ func (b *BlockChain) CalcNextRequiredDifficulty(timestamp time.Time) (uint32, er
...
+	return nil, nil
+}
+
+// CalcNextRequiredDifficultyFromValues computes the next required difficulty
+// from raw scalar values (height, bits, timestamps) rather than a full
+// HeaderCtx. Convenience wrapper for callers that maintain only lightweight
+// state (e.g. the headers presync state machine).
+//
+// prevTs should be -1 when no parent timestamp is available (e.g. genesis).
+func CalcNextRequiredDifficultyFromValues(params *chaincfg.Params, height int32,
```
jonathanMweiss: dead code? referenced by tests only.

node/blockchain/difficulty.go
```diff
@@ -222,3 +222,69 @@ func (b *BlockChain) CalcNextRequiredDifficulty(timestamp time.Time) (uint32, er
...
+func (c *valuesHeaderCtx) RelativeAncestorCtx(distance int32) chaincfg.HeaderCtx {
+	cur := c
+	for i := int32(0); i < distance; i++ {
+		if cur.parent == nil {
+			return nil
+		}
+		cur = cur.parent
+	}
+	return cur
+}
```
jonathanMweiss: `paramsChainCtx` and `valuesHeaderCtx` are stand-ins that only exist so presync can reach `calcNextRequiredDifficulty` / `CheckBlockHeaderContext`, and they return fake values (zero hash, grandparent `bits` 0, `VerifyCheckpoint` = `false`) for anything not reached today. 

I don't think this is the right way to go; adding mock-like code in production.
Perhaps the interface itself is the issue. 

WTEMA only needs `(params, lastBits, lastTs, parentTs)`, so could we extract that and the timestamp rule into value-based functions, make `calcNextRequiredDifficulty` / `CheckBlockHeaderContext` thin wrappers over them, and have presync call the value-based ones? That would remove `paramsChainCtx`, `valuesHeaderCtx`, `CalcNextRequiredDifficultyFromValues` (test-only today) and `CheckBlockHeaderContextFromValues`. 

If you'd prefer not to touch consensus code, we can look for another solution.

PR #112, opened 2026-05-20
updated vllm and cuda

raskin1:
## Summary

update vllm and cuda (to 13.0)


uv.lock
```diff
@@ -531,32 +531,15 @@ wheels = [
 
 [[package]]
 name = "cuda-bindings"
-version = "12.9.6"
+version = "13.1.1"
```
matan-dup: This is cuda-bindings 13.1.x, don't we want cuda-bindings 13.0?

matan-dup (dismissed):

uv.lock
raskin1: added a constraint

matan-dup (approved):

PR #125, opened 2026-05-28
fix(rpc): use chain.IsCurrent for getblocktemplate IBD check

ygd58:
getblocktemplate was using `SyncMgr.IsCurrent()` to check if the node is still in IBD. `Chain.IsCurrent()` is more accurate as it reflects the actual chain state directly.

One-line fix in `handleGetBlockTemplateRequest`.


ygd58:
Note: I noticed this same fix exists in the upstream branch fix/gbt-ibd-check. Opening this PR to help get it merged into master sooner if the upstream branch is not yet ready.

Aharonee (approved):

PR #133, opened 2026-05-31
fix(ci): use macos-13 runner for darwin-amd64 binary build

### LINKED ISSUE #127: Wrong binaries in pearl-wallet-v1.0.0 darwin-amd64 folder
## Component

https://github.com/pearl-research-labs/pearl/releases/tag/pearl-wallet-v1.0.0
- go-binaries-darwin-amd64-v1.0.2.tar.gz 

## Description

`go-binaries-darwin-amd64-v1.0.2.tar.gz` has **arm**64 binaries instead of **amd**64 binaries

## Steps to reproduce

1. cd go-binaries-darwin-amd64-v1.0.2
2. file ./oyster
3. Outputs the following:
./oyster: Mach-O 64-bit executable arm64

ygd58:
Fixes #127.

The `pearl-desktop-wallet.yml` workflow used `macos-15` for the macOS Intel (amd64) build. GitHub's `macos-15` runner runs on Apple Silicon hardware, so even with `GOARCH=amd64` the resulting binary is arm64 due to CGO and native toolchain.

`macos-13` is the last Intel-based macOS runner available on GitHub Actions and produces correct amd64 binaries.

### DIFF
--- a/.github/workflows/pearl-desktop-wallet.yml
+++ b/.github/workflows/pearl-desktop-wallet.yml
@@ -127,7 +127,7 @@ jobs:
             bin_ext: ""
             electron_build_cmd: pnpm -F @pearl/pearl-desktop-wallet run build:linux
           - name: macOS Intel
-            os: macos-15
+            os: macos-13
             goos: darwin
             goarch: amd64
             electron_arch: x64


ygd58:
Root cause: GitHub's macos-15 runner now runs on Apple Silicon hardware. Even with GOARCH=amd64 set, CGO and the native toolchain produce arm64 binaries on an arm64 host. macos-13 is the last Intel-based runner on GitHub Actions and will produce correct amd64 binaries.

This explains the behavior reported in #127 where go-binaries-darwin-amd64-v1.0.2.tar.gz contained an arm64 oyster binary.

PR #165, opened 2026-06-07
feat(zk-pow): Go FFI for PlainProof share verification and block proving

grandpooltech:
## Summary

Add Go FFI bindings so external tooling (e.g. a mining pool) can cheaply
verify the PlainProof "shares" miners submit and generate the ZK certificate
needed to assemble and submit a block.

## Type

feat

## Scope

zk-pow

## Changes

- `verify_plain_proof_ffi` — blake3-only PlainProof verification (no plonky2,
  no circuit cache); the cheap per-share check.
- `prove_plain_proof_ffi` — proves an existing PlainProof and returns
  `CZKProof{public_data, proof_blob}` so the caller can build a block
  certificate. Complements `prove_and_verify_plain_proof_ffi`, which only
  reports the proof length and discards the bytes.
- `mine_plain_proof` — CPU helper to produce a PlainProof (used by the tests).
- Export the new functions in `lib.rs`; add `bincode` dependency for PlainProof
  (de)serialization.
- Add unit tests: a mined PlainProof verifies (rc=0), a tampered one is
  rejected (rc≠0), and prove produces a non-empty certificate.

## Test plan

- [x] Existing tests pass (`task test`)
- [x] New tests added (`zk-pow/bindings/go/src/plain.rs`): verify accepts valid
      / rejects tampered; prove generates a certificate (the slow plonky2 case
      is `#[ignore]`d, run with `cargo test -- --ignored`)


o-k-d:
Thanks @grandpooltech.
@Aharonee WDYT?

Aharonee:
Seems like a reasonable addition, don't see a reason not to add it @o-k-d .

Should it be tested?

o-k-d:
@grandpooltech 
Can you add a (CI) test checking e.g. the following flow?
`miner -> plain_proof -> go_plain_verify -> go_prove -> go_zk_verify`

### Deserialization not panic-guarded in prove_plain_proof_ffi

verify_plain_proof_ffi correctly wraps deserialization inside catch_panic, but prove_plain_proof_ffi calls PlainProof::deserialize_compat outside of it (line 84). If that panics on malformed input, it crosses the FFI boundary -- which is UB. The existing mine.rs avoids this via the mine_and_prove helper. Please move the deserialization inside the catch_panic closure.

### Duplicated prove-and-copy boilerplate

The prove call + error handling (lines 92-103) and the copy-result-to-CZKProof block (lines 105-120) are near-identical to mine.rs's mine_and_prove and mine_inner. Consider extracting shared helpers into common.rs.

### Return code for null input in verify

verify_plain_proof_ffi returns 1 for null/empty input (line 31), but the docstring says 2 = bad input. The existing verify.rs returns 2 for null pointers. Should be 2 here too for consistency.

### Cargo.toml

The only change is removing a trailing newline -- no functional diff. Revert or drop from the PR.

grandpooltech:
@o-k-d done — panic-guarded the deserialize, extracted shared zk_prove/copy_prove_result into common.rs, fixed the null-input return code to 2, restored the Cargo.toml newline, and added the FFI flow test (plain_verify → prove → zk_verify) + a panic-guard regression test. Both green locally.

o-k-d (approved):

Aharonee (approved):

PR #179, opened 2026-06-12
Check deserialize().serialize()==id for proof objects

o-k-d:
Ensuring stark's salt is bound to blockheader::proofCommitment.


LindaGuiga (dismissed):
LGTM, thanks!

node/zkpow/verify.go
```diff
@@ -23,8 +23,17 @@ import (
 
 	"github.com/pearl-research-labs/pearl/node/chaincfg/chainhash"
 	"github.com/pearl-research-labs/pearl/node/wire"
+	"github.com/pearl-research-labs/pearl/version"
 )
 
+func init() {
+	got := C.GoString(C.zk_pow_ffi_version())
+	want := version.UserAgent()
```
Aharonee: This is fragile (it hand-syncs `bindings/go/Cargo.toml` with `version/version.go`, the single source of truth — a version bump there would panic every node) and redundant: staleness of the prebuilt `.a` belongs to the build toolchain, not a runtime check.

CI/Docker/release all go through `task build:blockchain` / `test:go`, whose targets `deps: [build:zk-gobind]`; that task runs `cargo build --release` unconditionally, so cargo rebuilds `libzk_pow_ffi.a` on any source change before `go build`.

Two suggestions:
- Add `deps: [build:zk-gobind]` to `unit` / `unit-race` / `unit-cover` / `integration` in `node/Taskfile.yml` — they run zkpow-tagged `go test` (which links the `.a`) without it, so `task unit-race` can link a stale lib.
- Then drop the `init()` check, `zk_pow_ffi_version()`, and the Cargo version bump.

zk-pow/src/api/proof_utils.rs
```diff
@@ -464,15 +474,18 @@ impl MiningConfiguration {
         let e = u16::from_le_bytes(trailer[0..2].try_into().unwrap());
         let top_k = u16::from_le_bytes(trailer[2..4].try_into().unwrap());
         ensure!(trailer[4..].iter().all(|&b| b == 0), "Reserved trailer bytes must be zero");
+        ensure!(e != 0 || top_k == 0, "top_k must be 0 when e == 0 (non-MoE)");
```
Aharonee, line 477: Should we add some short tests asserting this behavior?

o-k-d: I feel pretty confident with the `ensure_roundtrip` at the end of this function

node/zkpow/verify.go
o-k-d: Done

LindaGuiga (approved):

PR #187, opened 2026-06-14
perf(vllm-miner): cache the static-weight tensor_hash in the mining path

avifenesh:
Stack note: this PR is now stacked on #193 (direct int8 tensor_hash). The branch was rebased to preserve that order; after #193 lands, this diff should collapse to the static-weight hash cache layer.

## Modal H100 stacked benchmark

Environment: `NVIDIA H100 80GB HBM3`, CUDA `sm_90`, Torch `2.11.0+cu130`, repo Dockerfile image.

This isolates #187 on top of #193. BpEB caching is disabled (`cache_noised_weight=False`) so #194 does not affect the measurement. It compares full `pearl_gemm_noisy` with the weight-hash cache cleared before every call (simulating #193 without #187) against steady-state cache hits with the same weight and mining job.

| shape | no weight-hash cache | weight-hash cache hit | result |
|---|---:|---:|---:|
| gate_up 4096x28672x4096 | 1.797 ms | 1.700 ms | **+5.7%** |
| gate_up threshold 1024x28672x4096 | 1.115 ms | 1.013 ms | **+10.1%** |
| down 4096x4096x14336 | 1.271 ms | 1.180 ms | **+7.7%** |
| square 4096x4096x4096 | 0.781 ms | 0.757 ms | **+3.2%** |

## Summary
In mining mode the noisy GEMM seeds its proof-of-work commitment by keyed-merkle-hashing
BOTH operands (`tensor_hash`). Operand `B` is the layer **weight** — a persistent `nn.Parameter`
set once at load — and the hash key only changes when the mining job changes (gateway polled
~1/s), while a forward pass runs in milliseconds. So `tensor_hash(weight)` recomputes the
**identical** 32-byte root on every forward pass: provably redundant work.

This caches that root (shared helper `weight_hash_cache.cached_weight_hash`, used by both the
dense `pearl_gemm_noisy` and MoE `prepare_moe_noising` paths). The activation hash (`A`) is unchanged.

## Where the gain is: prefill-heavy mining
The noisy GEMM — and therefore the weight hash — only runs on GEMMs with **m ≥ 1024 tokens**
(the `should_use_noisy_gemm` gate). That is, **mining happens on heavy prefill, not on small
requests**: chunked-prefill steps cross 1024 tokens and hash; decode / small-request steps
(m = a few tokens) take the vanilla path and never hash. The cache's benefit scales with the
share of compute spent in noisy-prefill GEMMs.

This matters because a Pearl miner serves **many concurrent users with distinct contexts**, so
cross-request prefix-cache reuse is low and those large prefills are genuinely (re)computed — i.e.
they mine. The numbers below use distinct prompts (`--dataset-name random`), which models that
multi-user regime. (A single user replaying one growing conversation, where prefix caching reuses
the history, is the opposite case — prefills collapse, mining is rare, and the cache shows ~0%;
that is self-serve, not how Pearl serves.)

Measured on H100 / Llama-3.1-8B, mining ON, **interleaved** cache-vs-master A/B (alternated
run-by-run to remove run-order/thermal drift), call-count instrumentation for the noisy share:

| workload (distinct prompts) | noisy-GEMM share | cache vs master |
|---|---|---|
| prefill-saturated (in2048/out1, chunk 2048) | ~52% of GEMM calls | **+7.2% req/s** (15.05 vs 14.05) |
| heavy-prefill serving (in2048/out256) | ~14% of GEMM calls | **+3.1% req/s** (30.1 vs 29.2; +2.6% with prefix cache off) |
| multi-turn chat, full-context prefill (32×6 turns, no prefix reuse) | ~6% | **+2.5%** (wall 25.6 vs 26.3s) |
| decode-diluted (in64/out512) | ~1.6% | ~0% (mining barely runs) |

## Why it's correct (PoW-equivalent)
A cache **hit is always bit-identical** to recomputing, and every uncertainty degrades to a
redundant recompute — never a wrong hash:
- the mining-job **key bytes** are part of the cache key → a new job misses and recomputes;
- a **weakref** to the exact tensor is checked on a hit → a freed/reused `data_ptr` (ABA) misses;
- **device** type+index are in the key → CUDA `data_ptr` ints that collide across GPUs can't false-hit.

Valid under the immutable-weight invariant (Pearl weights are persistent params, never mutated
in place); `clear_cache()` is provided for any weight-reload path. All mining ops run on the
default CUDA stream, so the cached tensor's producer is ordered before its consumers — the same
same-stream assumption the un-cached code already relied on. Entries self-evict via a `weakref`
finalizer when the weight is GC'd.

## Tests
CPU unit test: a miss computes; a repeat hit returns the same object and does **not** recompute; a
key change recomputes; ABA (free a tensor, alloc a new one that may reuse the address) recomputes
rather than false-hitting; a GC'd weight auto-evicts its entry. Existing noisy-GEMM correctness
suite passes unchanged.

### DIFF
--- a/miner/pearl-gemm/csrc/tensor_hash/tensor_hash_api.hpp
+++ b/miner/pearl-gemm/csrc/tensor_hash/tensor_hash_api.hpp
@@ -54,6 +54,8 @@ void run_tensor_hash(
   TORCH_CHECK(key.dtype() == at::kByte, "key must be uint8");
   TORCH_CHECK(out.dtype() == at::kByte, "out must be uint8");
   TORCH_CHECK(roots.dtype() == at::kByte, "roots must be uint8");
+  TORCH_CHECK(data.dtype() == at::kByte || data.dtype() == at::kChar,
+              "data must be uint8 or int8");
   TORCH_CHECK(data.dim() == 2, "data must be 2D tensor");
   TORCH_CHECK(reinterpret_cast<uintptr_t>(data.data_ptr()) %
                       TMA_GLOBAL_ALIGNMENT_BYTES ==
@@ -103,8 +105,10 @@ void run_tensor_hash(
   auto stream = at::cuda::getCurrentCUDAStream();
   auto dprops = at::cuda::getCurrentDeviceProperties();
 
-  tensor_hash(data.data_ptr<uint8_t>(), data.numel(), out.data_ptr<uint8_t>(),
-              key.data_ptr<uint8_t>(), num_blocks,
+  // Hash the tensor storage bytes. int8 and uint8 are both one byte per
+  // element, so they must commit to identical roots for identical bit patterns.
+  tensor_hash(reinterpret_cast<uint8_t const*>(data.data_ptr()), data.numel(),
+              out.data_ptr<uint8_t>(), key.data_ptr<uint8_t>(), num_blocks,
               static_cast<uint32_t>(threads_per_block),
               static_cast<uint32_t>(num_stages),
               static_cast<uint32_t>(leaves_per_mt_block),
--- a/miner/pearl-gemm/src/pearl_gemm/pearl_gemm_interface.py
+++ b/miner/pearl-gemm/src/pearl_gemm/pearl_gemm_interface.py
@@ -559,7 +559,8 @@ def tensor_hash(
     """Hash a tensor using a key with configurable kernel parameters.
 
     Args:
-        data (Tensor): Input tensor to hash
+        data (Tensor): 2D uint8 or int8 tensor to hash as raw storage bytes.
+            int8 and uint8 inputs with the same bit pattern produce the same root.
         key (Tensor): Key tensor used for hashing
         out (Tensor): Output tensor for hash result
         roots (Tensor): Scratchpad tensor for intermediate results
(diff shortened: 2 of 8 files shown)


avifenesh:
Modal H100 stacked benchmark for #187 over #193.

This isolates the static-weight hash cache on top of the direct-int8 tensor_hash branch. BpEB caching was disabled (`cache_noised_weight=False`) so #194 does not contaminate the measurement. The benchmark compares full `pearl_gemm_noisy` with the weight-hash cache cleared before every call (simulating #193 without #187) against steady-state cache hits with the same weight and mining job.

| shape | no weight-hash cache | weight-hash cache hit | result |
|---|---:|---:|---:|
| gate_up 4096x28672x4096 | 1.797 ms | 1.700 ms | **+5.7%** |
| gate_up threshold 1024x28672x4096 | 1.115 ms | 1.013 ms | **+10.1%** |
| down 4096x4096x14336 | 1.271 ms | 1.180 ms | **+7.7%** |
| square 4096x4096x4096 | 0.781 ms | 0.757 ms | **+3.2%** |

This is kernel-path evidence for the stacked branch; the existing PR body still contains the older end-to-end serving benchmarks.

avifenesh:
Closing as superseded by #208 after rebasing onto #193.

#187 cached only the static B tensor hash. #208 keeps that benefit and extends the warm path to the full job-bound B-side mining state: `b_hash`, `commitment_b`, `EBL_*`, `EBR`, and `EBR_fp16`, with fresh H100 benchmark coverage on the #193-stacked branch.

The direct-int8 hash work remains in #193; #208 is stacked on top of it.

PR #193, opened 2026-06-16
perf(miner): hash int8 GEMM inputs without casts

avifenesh:
Stack layer 1/3.

This updates `tensor_hash` to accept int8 tensors directly and hash their raw storage bytes, so the dense mining path no longer has to stage A/B through `.to(torch.uint8)` before hashing.

## Modal H100 benchmark

Environment: `NVIDIA H100 80GB HBM3`, CUDA `sm_90`, Torch `2.11.0+cu130`, repo Dockerfile image.

This isolates #193 by comparing the new direct path (`tensor_hash(int8)`) against the old dense-mining behavior (`tensor_hash(t.to(torch.uint8))`) for hashing both A and B. The benchmark checks the two paths produce identical roots before timing.

| shape | old uint8-cast hash pair | direct int8 hash pair | result |
|---|---:|---:|---:|
| gate_up 4096x28672x4096 | 0.401 ms | 0.184 ms | **2.18x faster / +117.6%** |
| gate_up threshold 1024x28672x4096 | 0.380 ms | 0.182 ms | **2.09x faster / +108.9%** |
| down 4096x4096x14336 | 0.373 ms | 0.183 ms | **2.04x faster / +103.8%** |
| square 4096x4096x4096 | 0.140 ms | 0.080 ms | **1.74x faster / +74.1%** |

## Verification

- `uv run pytest tests/test_bpeb_cache.py tests/test_pearl_noisy_gemm.py::test_weight_hash_cache_hits_invalidates_and_is_aba_safe -q` run on the full stacked branch: 7 passed.
- Local CUDA tensor_hash focused tests could not run on this machine because the installed `pearl-gemm` extension is built for `sm_90a` while the local GPU is an RTX 5090 Laptop GPU, `sm_120`: no kernel image is available for execution on the device.

### DIFF
--- a/miner/vllm-miner/src/vllm_miner/gemm_operators.py
+++ b/miner/vllm-miner/src/vllm_miner/gemm_operators.py
@@ -123,7 +123,7 @@ def pearl_gemm_noisy(
 
     A_tensor_hash = torch.empty(32, device="cuda", dtype=torch.uint8)
     tensor_hash(
-        A.to(torch.uint8),
+        A,
         key_tensor,
         A_tensor_hash,
         tensor_hash_scratchpad,
@@ -131,7 +131,7 @@ def pearl_gemm_noisy(
 
     B_tensor_hash = torch.empty(32, device="cuda", dtype=torch.uint8)
     tensor_hash(
-        B.to(torch.uint8),
+        B,
         key_tensor,
         B_tensor_hash,
         tensor_hash_scratchpad,
--- a/miner/pearl-gemm/csrc/tensor_hash/tensor_hash_api.hpp
+++ b/miner/pearl-gemm/csrc/tensor_hash/tensor_hash_api.hpp
@@ -54,6 +54,8 @@ void run_tensor_hash(
   TORCH_CHECK(key.dtype() == at::kByte, "key must be uint8");
   TORCH_CHECK(out.dtype() == at::kByte, "out must be uint8");
   TORCH_CHECK(roots.dtype() == at::kByte, "roots must be uint8");
+  TORCH_CHECK(data.dtype() == at::kByte || data.dtype() == at::kChar,
+              "data must be uint8 or int8");
   TORCH_CHECK(data.dim() == 2, "data must be 2D tensor");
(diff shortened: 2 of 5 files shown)


miner/vllm-miner/src/vllm_miner/gemm_operators.py
```diff
@@ -123,15 +123,15 @@ def pearl_gemm_noisy(
 
     A_tensor_hash = torch.empty(32, device="cuda", dtype=torch.uint8)
     tensor_hash(
-        A.to(torch.uint8),
```
raskin1, line 126: i think you can just replace .to with .view without kernel changes etc

avifenesh: @raskin1 
yes, `.view(torch.uint8)` is the right idea correctness-wise. i measured it on H100, not only from torch semantics.

logic:
- `.to(torch.uint8)` copies.
- `.view(torch.uint8)` is a no-copy byte reinterpret, same data ptr, and it hashes the same bytes.
- old kernel + `.view` gets almost all the win vs `.to`.
- but it did not fully match this PR's direct-int8 path in the H100 hash-pair bench.

median ms on `NVIDIA H100 80GB HBM3`:

| shape | old `.to` | old kernel + `.view` | current direct int8 |
|---|---:|---:|---:|
| gate_up 4096x28672x4096 | 0.414 | 0.199 | 0.180 |
| gate_up threshold 1024x28672x4096 | 0.399 | 0.198 | 0.177 |
| down 4096x4096x14336 | 0.396 | 0.200 | 0.182 |
| square 4096x4096x4096 | 0.155 | 0.094 | 0.073 |

so i agree `.view` proves the no-copy part and is much better than `.to`, but i dont think we should shrink this PR to python-only `.view`. old-kernel `.view` was still ~9-22% slower than current direct int8 on the main hash-pair path.

also checked correctness: `.view` matched `.to`, matched python `MatrixMerkleTree` on the probe, and current direct int8 matched `.view`.

PR #197, opened 2026-06-17
Fuse B-side noising into GEMM mainloop when using spec decode

avifenesh:
## Summary

This PR introduces a new, opt-out-by-default way to “hide” the mining noise, along with a clean win of using EAGLE3.1 trained for the Pearl native model.
While the noise cant become cheap (or can but i didnt yet find how), it can be hiden in different placess.

The theory (that has been measured) behind it is that using spec decoder, the decod happen during speculation is memory bounded as any decode, but leave “free” compute.

First clean win is using the EAGLE3.1 - this is not final eagle recomend to end with, it was trained with 5 epoch and the winning epoch was epoch 3. For v1.1 the EAGLE should be trained more heavily with largest dataset. But this one was sufficient for a big win and fine acceptance.

As said, at least at this point i didnt found a good way to hide all the B noise.
Noising B still computes EARxBpEB not on chip, but as old way.
But Noising B skips the full BpEB HBM store and main GEMM forms BpEB on chip **during the spec decode** spec decode give us free compute. By doing BpEB during spec we hide a large fricion of the noise, and doing so on chip without HBM allow the spec decode to not get saturated. 
Important nuance - The spec decode is enogh to hide better **just** where k=4 or more, and thats require high acceptance rate. K=4 win in  some casses, and real full wins will require to get to good acceptance in k=5.
In the current trained eagle those cases are not the majority, and require more training for the EAGLE.

Following benchmarks where B in spec decode is matterialy better then vanila B noising in prefill, k=4, where k=4 win over other k=1..3.
Delta over vanila noising:
- short...long C1 = +~45%
- short C8 = +45%
- mid C8 = +45%
- long C8 = +18%
- Long C32: +16%
- Short C64: +13.4%
- Long C64: +9.1%

The results are, compare to vanila **no mining** is that from 1 up to 64C the PR method with mining perform **better**, while lowering the mining bound **down to 128**.  In average 20% higher throghput. 
Simple words, better results for clean inference, plus mining. Forget the noise.
In low C 
When getting to 64C the EAGLE and the vanila no mine get equal, but the EAGLE serving is still at least 10% better in throghput then vanila mining.

Benchmarks in comments.


avifenesh:
After rebasing on the caching B the gain is much lower/negative. so this pr is being closed while im researching other ways to make use of the spec time like pre-fettching the BpEB so it ready for the GEMM after the decode for A side. If any wont show improvments the EAGLE is by itself a good improvments to the throghput and will be contributed a standalone PR including the EAGLE itself.

PR #198, opened 2026-06-17
perf(miner): skip cold MoE expert mining

avifenesh:
## Summary

This adds an expert-local mining threshold for MoE GEMM1. The global prefill threshold still gates mining, but once it passes we now check the routed token count for each expert:

- hot experts continue to use Pearl noisy GEMM and submit proof work
- cold experts use vanilla int8 GEMM
- if no expert is hot, GEMM1 stays on the existing grouped int8 MoE path and skips noising setup

The old global-threshold behavior is still available with `MINER_MOE_EXPERT_LOCAL_MINING=0`.

## Why

Sparse MoE prefill can cross the global `M` threshold while each individual expert receives a much smaller slice. Mining every non-empty expert in that case adds noising/proof overhead to expert GEMMs that are too small to justify it.

## Validation

Modal H100 serving A/B on `pearl-ai/Qwen3-30B-A3B-Instruct-2507-pearl`, `input_len=2048`, `output_len=1`, 80 prompts, block submission disabled:

- Run 1: expert-local `7.603 req/s` vs global-old `5.997 req/s`, `+26.79%`
- Run 2: expert-local `7.773 req/s` vs global-old `6.228 req/s`, `+24.81%`

Modal smoke on H100:

- `test_moe_per_expert.py`: 3 passed
- `test_moe_block_submission.py::test_block_found_and_proof_verifies`: 1 passed

Local checks:

- `uv run --project miner/vllm-miner python -m py_compile miner/miner-base/src/miner_base/async_loop_manager.py miner/vllm-miner/src/vllm_miner/moe_gemm_operators.py miner/vllm-miner/src/vllm_miner/pearl_moe_experts.py miner/vllm-miner/tests/test_moe_block_submission.py modal_moe182_redirect_bench.py`
- `uvx ruff check miner/miner-base/src/miner_base/async_loop_manager.py miner/vllm-miner/src/vllm_miner/moe_gemm_operators.py miner/vllm-miner/src/vllm_miner/pearl_moe_experts.py miner/vllm-miner/tests/test_moe_block_submission.py modal_moe182_redirect_bench.py`
- `uvx ruff format --check --diff miner/miner-base/src/miner_base/async_loop_manager.py miner/vllm-miner/src/vllm_miner/moe_gemm_operators.py miner/vllm-miner/src/vllm_miner/pearl_moe_experts.py miner/vllm-miner/tests/test_moe_block_submission.py modal_moe182_redirect_bench.py`
- `git diff --check`

Modal runs:

[...]

PR #199, opened 2026-06-17
perf(miner): skip cold MoE expert mining on f8 path

avifenesh:
## Summary

Draft port of #198's expert-local MoE mining threshold onto `moe_f8_down_proj`.

This keeps the f8 branch's GEMM2/down-projection path intact and ports only the GEMM1 mining behavior:

- global prefill threshold still gates whether mining is considered
- per-expert routed token counts decide which experts actually run noisy GEMM
- cold experts use vanilla int8 GEMM for GEMM1
- if no expert crosses the local threshold, GEMM1 stays on the existing grouped int8 MoE path
- `MINER_MOE_EXPERT_LOCAL_MINING=0` restores the old global-threshold behavior

## Draft status

This is a code port, not yet a ready PR. The v1/all-int7 implementation has H100 smoke and serving evidence in #198. This f8 port still needs an H100 smoke and a serving benchmark on an fp8-down-compatible checkpoint.

The open performance question is mining density versus serving throughput: how many routed expert slots still get mined under expert-local threshold, and how that compares to the throughput lift. I am measuring that on the v1 benchmark path first, then we should repeat it on the f8 target once the right model/checkpoint is settled.

## Local validation

- `uv run --project miner/vllm-miner python -m py_compile miner/miner-base/src/miner_base/async_loop_manager.py miner/vllm-miner/src/vllm_miner/moe_gemm_operators.py miner/vllm-miner/src/vllm_miner/pearl_moe_experts.py miner/vllm-miner/tests/test_moe_block_submission.py`
- `uvx ruff check miner/miner-base/src/miner_base/async_loop_manager.py miner/vllm-miner/src/vllm_miner/moe_gemm_operators.py miner/vllm-miner/src/vllm_miner/pearl_moe_experts.py miner/vllm-miner/tests/test_moe_block_submission.py`
- `uvx ruff format --check --diff miner/miner-base/src/miner_base/async_loop_manager.py miner/vllm-miner/src/vllm_miner/moe_gemm_operators.py miner/vllm-miner/src/vllm_miner/pearl_moe_experts.py miner/vllm-miner/tests/test_moe_block_submission.py`
- `git diff --cached --check`

### DIFF
--- a/miner/miner-base/src/miner_base/async_loop_manager.py
+++ b/miner/miner-base/src/miner_base/async_loop_manager.py
@@ -233,6 +233,10 @@ class AsyncLoopManager:
             while not item.cuda_event.query():
                 await asyncio.sleep(0.01)
 
+            # Ensure host-pinned writes performed before the event are visible to
+            # the CPU thread before the callback reads the signal header.
+            item.cuda_event.synchronize()
+
             # callback should be fast (setup and call `handle_submit_block`)
             try:
                 item.callback(self.handle_submit_block)
--- a/miner/vllm-miner/src/vllm_miner/moe_gemm_operators.py
+++ b/miner/vllm-miner/src/vllm_miner/moe_gemm_operators.py
@@ -115,12 +115,40 @@ def _offsets_hash(
     return _hash_2d(offsets_bytes.reshape(1, -1), key_tensor, device)
 
 
+def build_moe_routing_layout(topk_ids: torch.Tensor, num_experts: int) -> MoERoutingLayout:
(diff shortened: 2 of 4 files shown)


avifenesh:
Closing this draft until it has f8-target H100 performance profiling. The code port exists locally on , but it should not be reviewed as a PR without perf evidence.

avifenesh:
Correction: the local f8 port branch is perf/moe188-expert-local-mining. Keeping this PR closed until we have f8-target H100 performance profiling, including mined-work density versus throughput.

PR #182, opened 2026-06-13
several bugfixes to make MoE mining consistent

raskin1:
## Summary

1. Buffer overflow in miner for small tensor hash
2. Using the right coherent mining_job in callback (test only issue)
## Type

bugfix

## Scope

tensor_hash and MoE miner

## Changes

- Now handling tensor_hash tail correctly

## Test plan

- [X] Existing tests pass (`task test`)
- [X] New small tensor hash tests

PR #188, opened 2026-06-14
feat(miner): switch quantization scheme of moe to f8 down proj

raskin1:
## Summary

feat: We change the quantization scheme for MoE models to use f8 block quantization in the down projection.

## Scope

MoE miner plugin

## Test plan

- [X] Existing tests pass (`task test`)

PR #201, opened 2026-06-18
perf(miner): skip cold MoE expert mining

avifenesh:
## Summary

This is the v1 MoE expert-local mining path on top of #182.
For MoE v2 - non public models exists yet, hence cant prove efficiency - cant open followup pr.

- Keep the existing global mining gate unchanged: mining is only attempted when the full MoE forward already passes `should_use_noisy_gemm(num_tokens, N, K)`.
- When that global gate passes, mine only experts whose routed-token count is large enough; cold experts use the vanilla int8 GEMM1 path.
- Keep the feature default-on, with an opt-out: `MINER_MOE_EXPERT_LOCAL_MINING=0` restores the old all-experts behavior after the global gate passes.
- Add `MINER_MOE_EXPERT_LOCAL_MIN_M` for serving-biased threshold experiments. Default is the global noisy-GEMM `min_m` (`1024` today). The override is intentionally constrained to `>=1024` and a multiple of `tile_size_m` (`128` today), so this PR does not mine below the current valid boundary.

## Benchmark evidence

Bench setup: RunPod H100, `pearl-ai/Qwen3-30B-A3B-Instruct-2507-pearl`, branch commit `829b00a3` for the expert-local behavior, `MINER_SKIP_BLOCK_SUBMISSION=1`, eager execution, no prefix cache. The threshold sweep used the same code path plus the threshold override that is now in this PR.

Baseline is the old/global MoE mining behavior: once the global gate passes, every expert in GEMM1 is mined. `mined slots/sec` combines serving throughput with mined routed slots/request, so it is the honest mining-throughput number.

| Shape | Threshold | Req/s vs old | Mined slots/request vs old | Mined slots/sec vs old | Read |
| --- | ---: | ---: | ---: | ---: | --- |
| input 1024, output 1, prompts 80 | 1024 | +37.1% | 88.2% | +21.0% | positive |
| input 2048, output 1, prompts 80 | 1024 | +29.2% | 87.3% | +12.7% | positive |
| input 3072, output 1, prompts 40 | 1024 | +29.4% | 87.9% | +13.7% | positive |
| input 2048, output 128, prompts 40 | 1024 | +12.9% | 88.3% | -0.3% | neutral/flat |
| input 2048, output 128, prompts 40 | 2048 | +16.9% | 77.0% | -10.0% | mining-negative |

No measured shape had negative serving throughput. The negative case is mined slots/sec: longer-output traffic spends more wall time outside the prefill-heavy mining window, so raising the expert threshold buys request throughput while losing effective mining throughput.

Threshold sweep summary:

| Shape | Best measured mining threshold | Mined slots/sec at 1024 | Mined slots/sec at 2048 | Serving delta at 2048 |
| --- | ---: | ---: | ---: | ---: |
| input 1024, output 1, prompts 80 | 1024 | +21.0% | +9.4% | +42.5% |
| input 2048, output 1, prompts 80 | 1024 | +12.7% | +4.8% | +39.8% |
| input 3072, output 1, prompts 40 | 1024 | +13.7% | +3.7% | +36.6% |
| input 2048, output 128, prompts 40 | 1024 | -0.3% | -10.0% | +16.9% |

Recommendation from these numbers:

- Keep `MINER_MOE_EXPERT_LOCAL_MIN_M` at `1024` when the goal is mined slots/sec.
- Raise the threshold only when serving throughput is the priority and lower mined slots/sec is acceptable.
- **For Pearl's stated usage, mining during prefill above 1024**, expert-local routing suggests the MoE threshold is probably too conservative for a future config/protocol experiment. This PR does not lower below 1024, but the data argues that if a lower valid MoE threshold is introduced later, it should be evaluated seriously because expert-local mining makes small cold experts cheap to skip.

## Validation

- `uv run pytest miner/vllm-miner/tests/test_moe_expert_threshold.py -q` - 12 passed
- `uv run pytest miner/vllm-miner/tests/test_vllm_interface.py::TestNoisyGemmSelectionThresholds -q` - 6 passed
- `uv run ruff check miner/vllm-miner/src/vllm_miner/pearl_moe_experts.py miner/vllm-miner/tests/test_moe_expert_threshold.py`
- `git diff --check`

### DIFF
--- a/miner/miner-base/src/miner_base/async_loop_manager.py
+++ b/miner/miner-base/src/miner_base/async_loop_manager.py
@@ -233,6 +233,10 @@ class AsyncLoopManager:
             while not item.cuda_event.query():
                 await asyncio.sleep(0.01)
 
+            # Ensure host-pinned writes performed before the event are visible to
+            # the CPU thread before the callback reads the signal header.
+            item.cuda_event.synchronize()
+
             # callback should be fast (setup and call `handle_submit_block`)
             try:
                 item.callback(self.handle_submit_block)
--- a/miner/vllm-miner/src/vllm_miner/moe_gemm_operators.py
+++ b/miner/vllm-miner/src/vllm_miner/moe_gemm_operators.py
@@ -115,12 +115,40 @@ def _offsets_hash(
     return _hash_2d(offsets_bytes.reshape(1, -1), key_tensor, device)
 
 
+def build_moe_routing_layout(topk_ids: torch.Tensor, num_experts: int) -> MoERoutingLayout:
+    """Build canonical routing data without committing or generating noise."""
+    top_k = topk_ids.shape[1]
+    num_routed_slots = topk_ids.shape[0] * top_k
+    device = topk_ids.device
+
+    routing_data = torch.empty(num_routed_slots, dtype=torch.int32, device=device)
+    slot_indices = torch.empty(num_routed_slots, dtype=torch.int32, device=device)
+    routing_offsets = torch.empty(num_experts, dtype=torch.int32, device=device)
+    routing_scratchpad = torch.empty(
+        get_build_routing_data_scratchpad_bytes(num_routed_slots),
+        dtype=torch.uint8,
+        device=device,
+    )
+    build_routing_data(
+        topk_ids.to(torch.int32).contiguous(),
+        routing_data,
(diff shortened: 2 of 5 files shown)


avifenesh:
Superseded by #206 after #188 (MoE v2) merged.

PR #205, opened 2026-06-18
feat(node,wallet): make log rotation size configurable via --logsize flag

s1ckoleg:
## Summary

Make the log file rotation size threshold configurable via a new `--logsize` flag for both `pearld` (node) and `oyster` (wallet), instead of being hardcoded at ~10MB.

## Type

feat

## Scope

node, wallet

## Changes

- `node/log.go` & `wallet/log.go`: `initLogRotator` now takes a `thresholdKB int64` parameter and passes it to `rotator.New`, replacing the hardcoded `10*1024`.
- `node/config.go` & `wallet/config.go`: add `defaultLogSizeKB = 10 * 1024`, a `LogSize uint32` config field (`--logsize`, also settable via the config file), default initialization, and validation rejecting `--logsize=0` (a 0 threshold would rotate on every write).
- `node/sample-pearld.conf` & `wallet/sample-oyster.conf`: document the new `logsize` option.

The default (10240 KB ≈ 10MB) preserves the previous behavior, so nothing changes unless the flag is set. Larger values let operators retain more log data per file before a rotation occurs. Negative values are rejected by `go-flags` at parse time because the field is unsigned.

## Test plan

- [ ] Existing tests pass (`task test`) — not run locally (requires the Rust/CUDA/C native toolchain); will run in CI.
- [ ] New tests added (if applicable) — N/A (config-flag threading; no new logic to unit-test)

Verified locally:
- `gofmt`/`goimports`/`go mod tidy`: clean, no extra changes.
- `go build ./node ./wallet` and `go vet ./node ./wallet`: pass.
- `go test ./node -run 'Config|Load'`: pass.


Aharonee (approved):
Looks good, thank you!

PR #236, opened 2026-07-06
ci: add pearld Docker image release and split into continuous build + promotion

Aharonee:
## Summary

Restructures the blockchain (pearld) build/release pipeline into a **continuous build** that publishes per-commit artifacts for internal/testnet use, plus a **promotion-only release** that ships the exact artifacts already built and tested — no rebuild. Adds a multi-arch `pearld` Docker image to GHCR along the way.

### Naming discipline
A version-shaped name is always an official release; a commit-hash name is always a build artifact. This makes it impossible to mistake a random commit's output for a release.

- Continuous (`blockchain_build.yml`): image `ghcr.io/.../pearld:sha-<commit>` only; binaries `pearl-<os>-<arch>-<sha>.<ext>` only.
- Release (`blockchain_release.yml`): image `:vX.Y.Z` + `:latest` only; binaries renamed to `pearl-<os>-<arch>-vX.Y.Z.<ext>` on the Releases page.

### `blockchain_build.yml` (new) — "Blockchain / Build & Publish"
- Triggers on `push: [master]` and `workflow_dispatch` (any branch, so a side branch can produce `sha-<commit>` images/binaries for internal + testnet testing).
- Builds the 5-platform binary matrix, bakes the commit into `pearld --version` via the `version.Build` ldflags hook, and uploads commit-hash-named workflow artifacts (30-day retention).
- Builds the multi-arch (`linux/amd64` + `linux/arm64`) image via Docker's official `docker/github-builder` reusable workflow (native per-arch runners, GHA cache, signed SLSA provenance + SBOM), tagged `sha-<commit>` only.

### `blockchain_release.yml` (reworked from the old `release.yml`) — "Blockchain / Release"
- No rebuild. Every run always promotes **both** the binaries and the image for the release commit (behind the `release` environment approval gate).
- Docker: re-tags the existing `sha-<commit>` manifest to `:vX.Y.Z` + `:latest` via `imagetools create` (preserves the multi-arch manifest + attestations). Fails clearly if the source image is missing.
- Binaries: downloads the tested artifacts from the `blockchain_build.yml` run for the commit (via `dawidd6/action-download-artifact`, SHA-pinned), renames them to clean version names, checksums, and creates the GitHub Release.

### Docs
- `CI_SECURITY_TIERS.md`: `blockchain_build.yml` documented as a Tier-2 continuous publisher (maintainer-triggered `packages: write`); `blockchain_release.yml` marked promotion-only.
- `node/docs/using_docker.md`: documents the tag naming discipline and points at the published GHCR image.

## Notes / trade-offs

- Image promotion is a pure re-tag; binaries are promoted by cross-run artifact download (truest "ship what you tested"). Artifact retention (30d) must exceed the commit-to-release gap.
- The `docker/github-builder` reusable workflow is pinned to `@v1` (matches Docker's docs); can be SHA-pinned if desired.
- The release publish is gated behind the `release` environment (approval), unlike the continuous build.
- Every master merge now runs the full binary matrix + image build; GHA cache keeps image builds cheap. A `paths` filter could be added later to skip doc-only merges.
- GHCR retention/cleanup of accumulating `sha-<commit>` versions is intentionally **deferred to a follow-up PR** (a safe implementation must be multi-arch/attestation aware).

## Test plan

- [ ] Merge to master (or dispatch `blockchain_build.yml` on a branch); confirm `pearld:sha-<commit>` and `pearl-*-<sha>` artifacts are produced.
- [ ] Dispatch `blockchain_release.yml` on that commit; confirm it pauses at the `release` approval, then re-tags the image to `:vX.Y.Z`/`:latest` and publishes renamed binaries.
- [ ] `docker buildx imagetools inspect ghcr.io/.../pearld:latest` lists both `linux/amd64` and `linux/arm64`.
- [ ] Release dispatched on a commit with no `blockchain_build.yml` run fails clearly (missing image / missing artifacts).


.github/workflows/release.yml
```diff
@@ -151,3 +156,32 @@ jobs:
...
+
+  docker:
+    needs: version
+    permissions:
+      contents: read
+      packages: write
+      id-token: write
+    uses: docker/github-builder/.github/workflows/build.yml@v1
+    with:
+      output: image
```
roiefink: Should we even run this if not pushing?

roiefink: Also will we ever want to release binaries but not docker image?

Aharonee: Good call — it no longer runs when not pushing. Dropped the build-only mode: both the new `docker-approval` gate and the `docker` job are now `if: ${{ inputs.push_image }}` / gated via `needs`, so unchecking `push_image` skips the image build entirely (`push`/`sbom` are now unconditionally true inside the job).

Aharonee: Yes — that's now supported: uncheck `push_image` on dispatch to release binaries only, and the docker job is skipped. Documented in the input description. Kept it coupled in this workflow rather than splitting into a separate one, per that toggle.

Aharonee:
Addressed the automated security findings:

- **release environment gate (High):** Fixed. Added a `docker-approval` job with `environment: release` that the `docker` job now `needs:`. Since a reusable-workflow caller can't set `environment:` directly, the gate lives in that dependency job, so the image publish requires the same release approval/branch policy as the binary release. Both jobs are also skipped entirely when `push_image=false`.
- **`@v1` mutable tag (Medium):** Acknowledged but intentionally keeping `@v1` for now to track Docker's maintained releases, matching their documented usage. Can revisit pinning to a full SHA if we want stricter supply-chain hygiene.

.github/workflows/release.yml
```diff
@@ -151,3 +156,40 @@ jobs:
...
         env:
           GITHUB_TOKEN: ${{ secrets.RELEASE_TOKEN }}
+
+  docker-approval:
+    needs: version
+    if: ${{ inputs.push_image }}
+    runs-on: ubuntu-latest
+    environment: release
+    steps:
+      - run: echo "Release environment approved; publishing pearld image."
```
roiefink: Do we actually need this? 

why not put 
`if: ${{ inputs.push_image }}` on the docker step itself?

PR #250, opened 2026-07-16
feat(wallet): use DNS seeders for peer discovery with optional custom peer

liad-madmon:
## Summary

Replaces the hardcoded `wallet-node0/1/2/3/4.pearlresearch.ai` peer addresses with the Pearl DNS seeders (`seeder1/2/3.pearlresearch.ai`) for automatic peer discovery, and adds an optional custom peer field in Settings for users who want to connect to a specific node.

## Type

feat

## Scope

wallet

## Changes

- Removed hardcoded `wallet-node0-4.pearlresearch.ai` mainnet peers and `node1-3.testnet.pearlresearch.ai` testnet peers; the oyster daemon now relies on its built-in DNS seeding (matching `chaincfg.Params.DNSSeeds`)
- `wallet-process.ts` only passes `--addpeer` when a custom peer is configured; otherwise the daemon falls back to DNS seeders automatically
- `peer-settings.ts` introduces `getCustomPeer()` returning `null` when no custom peer is set; legacy saved peers are detected and ignored for backwards compatibility
- `manager-service.ts` skips peer validation when no custom peer is configured (DNS seeders are self-validating)
- Redesigned `PeerSettingsModal.tsx` to show the DNS seeders as the default discovery method with an optional custom peer (IP/hostname + port) input
- Updated `app-bridge.ts` types and `network-config.ts` to reflect the new `dnsSeeders` field

## Test plan

- [ ] Existing tests pass (`task test`)
- [ ] Verified wallet launches and discovers peers via all 3 DNS seeders (25 peers each) in dev mode
- [ ] Verified custom peer field accepts IP/hostname + port and passes `--addpeer` to the daemon
- [ ] Verified clearing the custom peer field falls back to DNS seeders

### DIFF
--- a/apps/apps/pearl-desktop-wallet/src/main/config/consts.ts
+++ b/apps/apps/pearl-desktop-wallet/src/main/config/consts.ts
@@ -1,5 +1,32 @@
-export const MAINNET_DEFAULT_PEER_ADDRESSES = ['wallet-node0.pearlresearch.ai', 'wallet-node1.pearlresearch.ai', 'wallet-node2.pearlresearch.ai', 'wallet-node3.pearlresearch.ai', 'wallet-node4.pearlresearch.ai'];
-export const TESTNET_DEFAULT_PEER_ADDRESSES = ['node1.testnet.pearlresearch.ai', 'node2.testnet.pearlresearch.ai', 'node3.testnet.pearlresearch.ai'];
+// DNS seeders used by the oyster daemon for automatic peer discovery.
+// These match the DNSSeeds in node/chaincfg/params.go and are queried by the
+// node itself at startup — no --addpeer flag is needed.
+export const MAINNET_DNS_SEEDERS = [
+  'seeder1.pearlresearch.ai',
+  'seeder2.pearlresearch.ai',
+  'seeder3.pearlresearch.ai',
+];
+export const TESTNET_DNS_SEEDERS = [
+  'seeder1.testnet.pearlresearch.ai',
+  'seeder2.testnet.pearlresearch.ai',
+  'seeder3.testnet.pearlresearch.ai',
+];
+
+// Legacy hardcoded peer hosts from older wallet versions. Kept only so stale
+// peer-settings.json entries (saved as custom --addpeer targets) are detected
+// and ignored, falling back to DNS-seeder-based discovery.
+export const LEGACY_MAINNET_PEER_ADDRESSES = [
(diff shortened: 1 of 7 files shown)


Aharonee (commented):
The direction is right (no `--addpeer` by default, daemon discovers peers via its built-in DNS seeding), but a few things drift from the intent of the change — default to DNS seeders on startup, and let the user add custom peers to connect to:

**1. Don't bake the seeder hostnames into the wallet**
`MAINNET_DNS_SEEDERS` / `TESTNET_DNS_SEEDERS` are never passed to the daemon — they only feed the settings modal display. They duplicate the seed lists in `node/chaincfg/params.go` and will silently drift when those change. Suggest removing them entirely (along with the `dnsSeeders` field on `NetworkConfig`, `getPeerSettings()`, and the bridge type) and having the modal just say peers are discovered automatically, without enumerating hostnames. The `LEGACY_*` lists should stay — they drive the stale-settings migration.

**2. Support multiple custom peers, not a single one**
`--addpeer` is repeatable (`AddPeers []string` in `wallet/config.go`), and each entry becomes a permanent peer connected *in addition* to DNS-discovered ones. Since the goal is "allow the user to add custom peers", the setting should be a list: `customPeers: Array<{ address, port }>` in peer-settings, one `--addpeer` flag per entry in `getWalletArgs()`, and an add/remove list in the modal instead of a single address+port pair. The existing single-peer settings file can migrate on read the same way the legacy hardcoded hosts are already handled.

**3. A no-op save shouldn't kill the running wallet**
Save is now always enabled, and both `setCustomPeerAddress` and `resetPeerToDefault` unconditionally call `stopWalletProcess()`. Clicking Save with empty fields when no custom peer was configured stops the wallet session without changing anything. Only stop the wallet when the stored settings actually changed.

**4. Question: should an unreachable custom peer block startup?**
`selectWallet` still hard-fails when the custom peer doesn't validate. That made sense when the hardcoded peer was the only connection source, but custom peers are now additive — the daemon connects fine through DNS seeding without them. Consider downgrading this to a warning.

---

Verified against the daemon code: DNS seeding always runs in SPV mode regardless of `--addpeer` (`SeedFromDNS` in `spv/neutrino.go`, `DisableDNSSeed` defaults to false), and repeated `--addpeer` values each become permanent peers alongside discovered ones — so multiple custom peers are purely additive with no discovery trade-off.

Aharonee (approved):

PR #275, opened 2026-08-05
feat(consensus): rank-penalty softfork

Aharonee:
## Summary

A proof's jackpot bound is scaled by the work one attempt costs, and the noise rank divides that work, so mining at a larger rank buys a proportionally easier bound for the same effort. From `RankPenaltyForkHeight` on, a certificate's noise rank must be at least 128 and its jackpot must meet a bound scaled by `128/rank`, which cancels the advantage. The rule only tightens validity, so pre-fork nodes accept every post-fork block.

## Type

feat

## Scope

node (blockchain, chaincfg, zkpow, mining, netsync, wire), spv, zk-pow, py-pearl-mining, miner

## Changes

- **The rule** (`zk-pow/src/api/sanity_checks.rs`): `PENALTY_BASE_RANK = 128`, `check_rank_penalty`, and `penalized_target_bound`. The consensus path still saturates at `U256::MAX`, where an absurdly easy nbits legitimately means every jackpot passes; `penalized_target_bound` returns `None` instead, because a miner scaling a share target must be told the target is unusable rather than handed one every hash satisfies.
- **Wire parsing** (`zk-pow/src/api/proof_utils.rs`): named byte ranges replace the open-coded offsets, plus `mining_config_and_jackpot_from_wire_bytes`, which reads only the prefix dense and MoE proofs share so the rule needs neither the MoE tail nor a block header.
- **FFI**: `check_rank_penalty` and `default_mining_config` entry points, and an exported `MIN_NOISE_RANK` asserted at compile time against `PENALTY_BASE_RANK`.
- **Certificate rules** (`node/blockchain/validate.go`): `CheckCertificateVersion` becomes `CheckCertificateRules`, enforcing the version cutover, the dense-only fork and the rank-penalty rule together, and taking the header and the behavior flags the new rule reads. Adds `ErrRankPenalty`, and a `blockchain.New` assertion that the fork cannot precede `MoEForkHeight` since only V2 certificates carry a noise rank.
- **`NetBehaviorFlags`** (`node/blockchain/process.go`): derives the `BFNoPoWCheck` that SimNet needs in one place, replacing the three copies of that special case in `checkBlockSanity`, `CheckHeaderSanity` and the SPV block manager. The rank-penalty rule reads a proof, so it is skipped wherever proof of work is not verified, which includes block templates.
- **`CertificateV2.PublicDataBytes()`**: the `PublicData`/`PublicDataLen` pairing now appears once. `verifyCertificateV2`'s inline guard against an out-of-range length is gone: `Deserialize`, `Mine` and `MineMoE` are the only producers and all three bound it, and `ProofCommitment` sliced the same fields unguarded regardless.
- **Node miner** (`node/zkpow/miner.go`): the serialized mining configuration now comes from Rust's `default_mining_config` rather than a hardcoded byte array, and sits at rank 128 — the floor, so mined blocks stay valid past activation without paying a penalty. `DefaultNoiseRank` and `DefaultMMAType` had no other readers and are dropped.
- **Miner gateway**: `MiningJob.adjust_target` goes through `penalized_target_bound` and raises on a sub-floor rank, a degenerate configuration, or an unrepresentable bound; `MinerSettings.noise_rank` rejects values below the floor at load time rather than mining blocks consensus will reject.
- **Activation heights** (`node/chaincfg/params.go`): mainnet 96251, testnet 36761, testnet2 80627, and regtest 1, which verifies proofs.
- **Version**: 1.2.2 -> 1.3.0.

## Test plan

- [x] New tests added
  - `node/zkpow/rank_penalty_test.go` (zkpow tag) pins the Rust rule to an independent Go implementation of the same formula, and covers the rank floor plus public data the rule cannot parse. Testing each rank at its own bound also covers the property the fork exists for: a jackpot that wins at the base rank is rejected at a larger one.
  - `node/blockchain/rank_penalty_fork_test.go` covers when the rule applies: the activation boundary, the `BFNoPoWCheck` skip, block templates, and the params ordering invariant.
  - `node/mining/mining_test.go` now builds a template on regtest, which verifies proofs, so the template path is exercised with the certificate forks active; the existing coverage was SimNet-only, where the rule is skipped for the network regardless.
  - Rust unit tests for neutrality at the base rank, the proportional penalty above it, the inclusive bound, degenerate configurations, and unrepresentable bounds; Python tests for `adjust_target`.
- [x] Ran locally: `go build ./...`, `go test` for `node/blockchain`, `node/chaincfg`, `node/mining`, `node/wire`, `node/netsync` and `spv`; `gofmt`, `goimports` and `go mod tidy` clean; `cargo fmt --check` on `zk-pow`, `zk-pow/bindings/go` and `py-pearl-mining`; `ruff check` and `ruff format --check` on the changed Python files. `TestCheckBlockSanity` and the `node/wire` certificate tests fail identically on master without `-tags zkpow`.
- [ ] Full `task test` and the zk-pow and py-pearl-mining suites left to CI.

## Rollout

The testnet heights were rescheduled in the second commit, projected from the tips at each chain's own recent spacing (both run slow against the 3m14s target):

- testnet 36761, 7 blocks from tip 36754 at ~493 s/block, so ~19:15 IDT today
- testnet2 80627, 30 blocks from tip 80597 at ~243 s/block, so ~20:20 IDT today
- mainnet 96251, unchanged, ~283 blocks from tip 95968, so ~08:45 IDT tomorrow

Testnet stays roughly an hour ahead of testnet2 so problems surface on the internal chain first. Nodes must be upgraded before their network's height: an activation height that passes before the rollout completes would have upgraded nodes reject the blocks mined after it, so any height at risk needs rescheduling first.

### DIFF
--- a/miner/conftest.py
+++ b/miner/conftest.py
@@ -21,7 +21,7 @@ from pearl_gateway.comm.dataclasses import (
     OpenedBlockInfo,
 )
 from pearl_gateway.config import MinerRpcConfig, PearlConfig
-from pearl_mining import MatrixMerkleProof, MerkleProof, PlainProof, ZKProof
+from pearl_mining import MatrixMerkleProof, MerkleProof, PlainProof, ZKProof, penalized_target_bound
 
 
 @pytest.fixture(scope="session")
@@ -343,7 +343,10 @@ def _adjust_mining_job_target(
     mining_config: MiningConfiguration,
     target: int,
 ) -> MiningJob:
-    target //= MiningJob._get_difficulty_adjustment_factor(mining_config)
+    bound = penalized_target_bound(1, mining_config)
+    if bound is None:
+        raise ValueError("degenerate mining config in test scaffolding")
+    target //= bound
     return MiningJob(incomplete_header_bytes=incomplete_header_bytes, target=target)
 
 
--- a/miner/miner-base/src/miner_base/settings.py
+++ b/miner/miner-base/src/miner_base/settings.py
@@ -1,3 +1,5 @@
+from pearl_mining import PENALTY_BASE_RANK
+from pydantic import field_validator
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
 
@@ -34,3 +36,13 @@ class MinerSettings(BaseSettings):
     no_vllm_plugin: bool = False
 
     enable_async_cuda_event_processing: bool = True
+
+    @field_validator("noise_rank")
+    @classmethod
+    def _noise_rank_at_least_base(cls, noise_rank: int) -> int:
+        if noise_rank < PENALTY_BASE_RANK:
+            raise ValueError(
+                f"noise_rank {noise_rank} is below the minimum {PENALTY_BASE_RANK}; "
+                f"blocks would be rejected by consensus"
+            )
+        return noise_rank
--- a/miner/pearl-gateway/src/pearl_gateway/comm/dataclasses.py
+++ b/miner/pearl-gateway/src/pearl_gateway/comm/dataclasses.py
@@ -19,7 +19,7 @@ from pearl_gateway.comm.mining_configuration import (
 from pearl_gateway.rpc_types import (
     GetBlockTemplateResponse,
 )
-from pearl_mining import IncompleteBlockHeader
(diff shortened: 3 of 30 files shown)


raskin1 (dismissed):

raskin1 (approved):

Boinsss (commented):
H

PR #280, opened 2026-08-11
feat(consensus): salted noise-seed hardfork

Aharonee:
## Summary

- **Salted Noise-Seed Derivation**: Updates Blake3 commitment hash calculations to use domain-separated salted hashing for noise-seed derivation (`hash_a` and `hash_b`).
- **Certificate Version V3 & Protocol Version 2**: Introduces `wire.CertificateVersionV3`, bumps `wire.ProtocolVersion` to 2, and enforces strict version cutovers in block validation logic.
- **Cross-Component Alignment**: Updates CUDA GEMM kernels in `pearl-gemm`, `zk-pow` Rust library & FFI bindings, `py-pearl-mining`, and gateway/miner RPC schemas.
- **Activation Heights**: Configured activation heights across networks: MainNet (`98900`), TestNet (`38648`), TestNet2 (`83109`), and active from genesis on RegTest & SimNet (`1`).
- **Documentation & Tests**: Includes node and library unit tests, pinned vector assertions, and upgrade documentation in `docs/salted-seed-fork-upgrade-guide.md`.

## Test plan

- [x] Verified Go unit tests in `node/chaincfg`, `node/blockchain`, `node/peer`, `node/wire`, and `node/zkpow`
- [x] Verified Rust unit tests and pinned vector checks in `zk-pow`
- [x] Verified clean rebase and zero conflicts against `master`

PR #281, opened 2026-08-11
fix(zkpow): pin Merkle tree leaf counts to declared dimensions

raskin1:
## Summary

Follow-up to the salted noise-seed hardfork (#280). That PR made the declared `m`/`n` consensus-critical by salting the V3 noise seed with them (`bind_root_a`/`bind_root_b`), but `PlainProof::parse_proof` never validated that a proof's committed Merkle trees actually have the leaf count those dimensions imply. `MerkleProof::total_leaves` was carried in the proof and, for the A/B trees, never checked against `m`/`n`/`k` at all.

This adds `check_declared_tree_sizes`, called at the top of `parse_proof`, which rejects any plain proof whose:
- **A** tree doesn't declare `padded_chunk_len(m*k)/CHUNK_LEN` leaves,
- **B^T** tree doesn't declare `padded_chunk_len(total_b_cols*k)/CHUNK_LEN` leaves (`total_b_cols = n` dense, `n*e` MoE), or
- **MoE routing** tree doesn't declare `padded_chunk_len(m*top_k*4)/CHUNK_LEN` leaves.

This keeps a miner from opening one committed "carousel" tree under several dimension interpretations. The salted seed already gives each interpretation a distinct seed; this makes the dimension↔tree binding explicit and local rather than emergent. Dimension products use checked arithmetic so adversarial inputs error instead of overflowing.

Honest proofs are unaffected: their trees are built over exactly `padded_chunk_len(bytes)` leaves, which is the same size `compute_external_cvs` already uses to reconstruct the roots.

## Test plan

- [x] `cargo test --lib ffi::plain_proof` — 11 passed, including two new tests covering dense off-by-one rejection and the MoE routing-tree leaf count.
- [x] `rustfmt --check` clean.
- [ ] CI: full `zk-pow` suite (incl. `moe_test` end-to-end mine/parse/verify) to confirm honest proofs still verify.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/zk-pow/src/ffi/plain_proof.rs
+++ b/zk-pow/src/ffi/plain_proof.rs
@@ -413,6 +413,63 @@ impl PlainProof {
         if let Some(moe) = &self.moe { self.n * moe.e } else { self.n }
     }
 
+    /// Leaf count the Merkle tree of a row-major byte buffer whose length is the
+    /// product of `dims` must declare. Errors if the product overflows `usize`.
+    fn expected_merkle_leaves(dims: &[usize]) -> Result<usize> {
+        let bytes = dims
+            .iter()
+            .try_fold(1usize, |acc, &d| acc.checked_mul(d))
+            .ok_or_else(|| anyhow::anyhow!("declared dimensions {dims:?} overflow usize"))?;
+        Ok(pearl_blake3::padded_chunk_len(bytes) / BLAKE3_CHUNK_LEN)
+    }
+
+    /// Rejects a proof whose committed Merkle trees don't have the leaf count
+    /// implied by the declared dimensions.
+    ///
+    /// The V3 noise seed is salted with the declared `m`/`n`, making them
+    /// consensus-critical: without pinning each tree's `total_leaves` to those
+    /// dimensions a miner could open one committed tree under several dimension
+    /// interpretations. The A/B trees are row-major `m*k` / `total_b_cols*k`
+    /// int8 bytes; the MoE routing tree is `m*top_k` little-endian `u32`s.
+    fn check_declared_tree_sizes(&self) -> Result<()> {
+        let a_expected = Self::expected_merkle_leaves(&[self.m, self.k])?;
+        ensure_eq!(
+            self.a.proof.total_leaves,
+            a_expected,
+            "A Merkle tree declares {} leaves but m={} k={} imply {}",
+            self.a.proof.total_leaves,
+            self.m,
+            self.k,
+            a_expected
+        );
+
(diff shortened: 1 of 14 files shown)


o-k-d (dismissed):

raskin1:
**Why the vLLM Docker check is red (and will stay red on this PR):**

The `Test Building Docker image (vLLM miner)` job calls `docker/build-push-action` without a `context:` input, so it builds from the Git context — which on a `pull_request_target` event is the **base branch**, not the PR. It has been compiling `master`'s `commitment_hash_from_merkle_roots_kernel.hpp:180` (broken since #280 merged: nvcc rejects the ODR-use of `blake3::CHAINING_VALUE_SIZE_U32` in device code) on every PR, regardless of PR content.

This PR carries both fixes:

- `6c10573` fixes the kernel (index via cute `Int<>`, a pure constant-expression context). GPU CI and Integration Tests — which *do* build the PR checkout — compile it and pass.
- `71ce978` adds `context: .` to the workflow so the job builds the approved checkout going forward (same as `miner_docker_release.yml` already does). Note this takes effect only after merge, since `pull_request_target` runs the base branch's workflow definition.

So the Docker check self-heals for all PRs once this merges; it cannot go green on this PR itself.

matan-dup (dismissed):

Aharonee (approved):

zk-pow/src/ffi/plain_proof.rs
```diff
@@ -413,6 +413,63 @@ impl PlainProof {
...
+
+    /// Rejects a proof whose committed Merkle trees don't have the leaf count
+    /// implied by the declared dimensions.
+    ///
+    /// The V3 noise seed is salted with the declared `m`/`n`, making them
+    /// consensus-critical: without pinning each tree's `total_leaves` to those
+    /// dimensions a miner could open one committed tree under several dimension
+    /// interpretations. The A/B trees are row-major `m*k` / `total_b_cols*k`
+    /// int8 bytes; the MoE routing tree is `m*top_k` little-endian `u32`s.
+    fn check_declared_tree_sizes(&self) -> Result<()> {
```
o-k-d, line 434: Perhaps just remove `total_leaves`? It seems to not being used in an interesting way.
Instead, here we can check that the merkle proof of the declared opened leaves match the shape one would expect.
Alternatively, fill/override here the total_leaves field and later check the proof shape agrees.
Not a strong opinion.

o-k-d (approved):

zk-pow/src/ffi/plain_proof.rs
raskin1: its a bigger change.. which will reuiqre a miner change also

PR #295, opened 2026-08-24
fix(peer): preserve addresses in PushAddrV2Msg shuffle

### LINKED ISSUE #292: bug(peer): PushAddrV2Msg shuffle duplicates addresses
## Problem

`node/peer.(*Peer).PushAddrV2Msg` corrupts the address list when it contains more than `wire.MaxV2AddrPerMsg` entries. Its `rand.Shuffle` callback performs the swap as two sequential assignments:

```go
m.AddrList[i] = m.AddrList[j]
m.AddrList[j] = m.AddrList[i]
```

The first assignment overwrites the original value at `i`, so the second assignment copies the same value back to `j`. Repeated swaps introduce duplicate addresses and discard others before the list is truncated.

This affects large `addrv2` responses to `getaddr`, including responses consumed by DNS seed crawlers.

## Expected behavior

Shuffling should preserve the input as a permutation before truncation:

```go
m.AddrList[i], m.AddrList[j] = m.AddrList[j], m.AddrList[i]
```

Add a regression test with more than `wire.MaxV2AddrPerMsg` distinct addresses that verifies the returned list has the expected size and contains no duplicates.

jonathanMweiss:
## Summary

- `node/peer.(*Peer).PushAddrV2Msg` shuffled addresses with two sequential assignments, so the first write clobbered the value at `i` before the second read it back. Repeated swaps introduced duplicate addresses and dropped others before the list was truncated to `wire.MaxV2AddrPerMsg`.
- Replaced the broken swap with a simultaneous swap (`a[i], a[j] = a[j], a[i]`), making the shuffle a true permutation of the input — matching the pattern already used by `PushAddrMsg`.
- Added a regression test (`TestPushAddrV2MsgShuffleNoDuplicates`) that feeds more than `wire.MaxV2AddrPerMsg` distinct addresses and asserts the returned list is capped to the maximum, contains no duplicates, and every returned address was part of the original input. Verified the test fails on the buggy code (`duplicate address "10.0.0.0"`) and passes with the fix.

This affects large `addrv2` responses to `getaddr`, including responses consumed by DNS seed crawlers.

Fixes #292.

## Test plan

- [x] `go test ./node/peer/ -run TestPushAddrV2MsgShuffleNoDuplicates -count=1` passes with the fix
- [x] Confirmed the test fails on the pre-fix code (catches the duplicate)
- [x] `go vet ./node/peer/` clean
- [x] `gofmt` clean
- [x] Full `node/peer` test suite passes

Made with [Cursor](https://cursor.com)


Aharonee (commented):
LGTM — the swap fix is correct and the regression test catches the bug (verified fail-before/pass-after). One non-blocking suggestion posted inline: the test could be simplified by dropping the connection scaffolding, since the bug is in the pure shuffle logic. Approving; the simplification is optional.

Aharonee (dismissed):
LGTM — the tuple-swap fix is correct and the regression test catches the bug (verified fail-before/pass-after). Non-blocking suggestion posted inline to simplify the test; optional.

node/peer/peer_test.go
```diff
@@ -644,3 +644,73 @@ func TestSendAddrV2Handshake(t *testing.T) {
 	inPeer.WaitForDisconnect()
 	outPeer.WaitForDisconnect()
 }
+
+// TestPushAddrV2MsgShuffleNoDuplicates is a regression test for issue #292.
+// It verifies that PushAddrV2Msg preserves the address set as a permutation
+// when the input contains more than wire.MaxV2AddrPerMsg distinct addresses:
+// the returned list is capped to MaxV2AddrPerMsg, contains no duplicates, and
+// every returned address was part of the original input.
+func TestPushAddrV2MsgShuffleNoDuplicates(t *testing.T) {
```
Aharonee, line 653: Consider this lighter version of a test instead:

```golang
func TestPushAddrV2MsgShuffleNoDuplicates(t *testing.T) {
	const numAddrs = wire.MaxV2AddrPerMsg + 250

	peerCfg := &peer.Config{
		UserAgentName:    "peer",
		UserAgentVersion: "1.0",
	}
	p, err := peer.NewOutboundPeer(peerCfg, "10.0.0.1:8333")
	require.NoError(t, err)

	addrs := make([]*wire.NetAddressV2, numAddrs)
	original := make(map[string]struct{}, numAddrs)
	for i := range addrs {
		ip := net.IPv4(10, 0, byte(i>>8), byte(i))
		addrs[i] = wire.NetAddressV2FromBytes(
			time.Unix(int64(i), 0), 0, ip.To4(), 8333,
		)
		original[addrs[i].Addr.String()] = struct{}{}
	}

	sent, err := p.PushAddrV2Msg(addrs)
	require.NoError(t, err)
	require.Len(t, sent, wire.MaxV2AddrPerMsg)

	seen := make(map[string]struct{}, len(sent))
	for _, a := range sent {
		key := a.Addr.String()
		_, dup := seen[key]
		require.Falsef(t, dup, "duplicate address %q", key)
		_, ok := original[key]
		require.Truef(t, ok, "address %q was not in the input set", key)
		seen[key] = struct{}{}
	}
}
```

Aharonee (approved):

PR #298, opened 2026-08-27
fix(txscript): bound HashCache and SigCache with random eviction

jonathanMweiss:
## Summary

Fixes GHSA-gpqh-6w3v-xhhv — unbounded `HashCache` allowing memory exhaustion via peer transaction spam.

### Root cause
`HashCache` and `SigCache` grew without bound. An attacker could flood the node with witness transactions, each inserting a unique sighash midstate, causing unbounded memory growth.

### Fix
- Add `randEvictCache` — a generic bounded cache with random eviction shared by both `SigCache` and `HashCache`.
- `NewHashCache(maxSize)` / `NewSigCache(maxSize)` now enforce a hard cap; `maxSize == 0` is a true no-op cache.
- Default size is 100,000 entries (configurable via `--sigcachemaxsize`), shared by both caches.

### Test plan
- [x] `go test -tags zkpow ./node/blockchain/ ./node/txscript/ ./node/mining/ ./node/mempool/`
- [x] `go build ./...` + `go vet ./...`
- [x] `TestHashCacheEviction` / `TestHashCacheDisabled` verify the cache stays bounded
- [x] `TestCheckBlockSanityRejectsMultipleCoinbases` runs without `-tags zkpow` via `BFNoPoWCheck`

### Version
Bumps to 1.4.6.

### Credit
Reported by [@t-allfather](https://github.com/t-allfather). Thank you for the responsible disclosure.

Refs GHSA-gpqh-6w3v-xhhv


Aharonee (commented):
Two notes on `TestCheckBlockSanityRejectsMultipleCoinbases`, both non-blocking. The rest of the change set looks good to me — the cache bounding, the `IsCoinBase` skip, and the required-cache contract from f6a4506 all read well.

node/blockchain/validate_test.go
```diff
@@ -195,6 +195,57 @@ func TestCheckBlockSanity(t *testing.T) {
...
+	chainParams := &chaincfg.RegressionNetParams
+	header := wire.BlockHeader{
+		Version:    1,
+		PrevBlock:  chainhash.Hash{},
+		MerkleRoot: Block100000.BlockHeader().MerkleRoot,
+		Timestamp:  time.Unix(time.Now().Unix(), 0),
+		Bits:       chainParams.PowLimitBits,
+	}
+
+	cert, err := zkpow.Mine(&header, chainParams.RequiredCertVersion(1))
```
Aharonee: The ZK proving here is out of scope for what this test verifies. The multiple-coinbase rule is context-free and network-independent — a plain loop over `transactions[1:]` with no `chainParams` involvement — yet mining drags the `zkpow` build tag and the native Rust FFI bindings into the one test that guards the invariant `checkBlockScripts` now relies on. Without `-tags zkpow` (and a built `zk_pow_ffi.h`) it silently doesn't run at all, and it roughly doubles the package's proving time; the neighbouring `TestCheckBlockSanity` concedes proving is slow in its own comment.

It can't just be deleted, though: `checkBlockSanity` runs `CheckBlockHeaderSanity` -> `checkProofOfWork` before it ever reaches the coinbase loop, so on RegTest an unmined certificate fails earlier and the `ErrMultipleCoinbases` assertion is never reached. Two ways out, since this file is `package blockchain`:

- Call the unexported `checkBlockSanity(block, chainParams, timeSource, BFNoPoWCheck)` directly. `BFNoPoWCheck` turns `checkProofOfWork` into a no-op; the exported `CheckBlockSanity` hardcodes `BFNone`, which is the only reason real proving is forced here.
- Or use `chaincfg.SimNetParams`, which auto-skips PoW (per the comment in `TestCheckBlockSanity`).

Either way a stub certificate is probably still needed to satisfy the header's structural checks — genesis uses a vanity `0xdeadbeef` proof — so this is swapping the mined cert for a cheap one rather than dropping it.

node/blockchain/validate_test.go
```diff
@@ -195,6 +195,57 @@ func TestCheckBlockSanity(t *testing.T) {
...
+	}
+	block := btcutil.NewBlock(msgBlock)
+
+	timeSource := NewMedianTime()
+	err = CheckBlockSanity(block, chainParams, timeSource)
+	if err == nil {
+		t.Fatalf("CheckBlockSanity accepted a block with two coinbases")
+	}
+
+	rerr, ok := err.(RuleError)
```
Aharonee: This test uses raw `t.Fatalf` throughout, while the rest of the package has moved to testify (`chain_test.go`, `process_test.go`, `merkle_test.go`, `difficulty_test.go` and others) — and the other test added in this PR, `node/mempool/hashcache_spam_test.go`, uses `require` throughout, so the PR is internally inconsistent. `validate_test.go` being an old btcd-derived holdout with no testify usages is a good reason not to churn the tests already in it, but new ones should follow the newer convention.

Concretely, this block reads better and gets more robust as:

```go
var rerr RuleError
require.ErrorAs(t, err, &rerr)
require.Equal(t, ErrMultipleCoinbases, rerr.ErrorCode)
```

`ErrorAs` also survives the error being wrapped somewhere down the line, which the bare type assertion would not.

Two nits while here: the `t.Fatalf` on L236 has no format arguments, so it should be `t.Fatal`, and this test is missing `t.Parallel()` even though every txscript test added in this PR has it.

Separately, same test (minor): the doc comment says `checkBlockScripts` "skips Transactions()[0] as the only coinbase", but the code calls `IsCoinBase(tx)` on every transaction in the loop, not just index 0.

node/mempool/hashcache_spam_test.go
```diff
@@ -0,0 +1,130 @@
...
+// sighash midstate of every witness transaction that makes it as far as script
+// validation. Nothing purges that entry when validation then fails, and the
+// sync manager only suppresses repeats of a txid it has already rejected, so an
+// attacker who mutates a txid-committed field on each attempt gets a fresh
+// cache key every time. A single spendable witness output is enough to do this
+// indefinitely.
+//
+// The test asserts the midstate really is cached on the rejected path (so it is
+// exercising the vulnerable insertion) while the cache stays within its configured maximum.
+func TestHashCacheBoundedUnderPeerSpam(t *testing.T) {
```
Aharonee: Suggest dropping this file. The bound is already covered by `TestHashCacheEviction` and `TestHashCacheDisabled` in `node/txscript`, and the two things this adds on top don't hold up:

- **It isn't a peer-level test.** It calls `ProcessTransaction` directly, so `OnTx`, the sync manager, and the bounded rejected-txid set — the netsync machinery that makes distinct txids matter in the advisory — are never exercised. The `attackerPeerID` tag is inert.
- **It doesn't cover production wiring.** L49 swaps in a locally constructed cache (`harness.txPool.cfg.HashCache = hashCache`), so the sizing that actually protects a node — `server.go` passing `cfg.SigCacheMaxSize` to both caches — stays untested either way. That was the main thing I'd expect an end-to-end test here to pin down.

jonathanMweiss: Dropped it. The bound is already covered by `TestHashCacheEviction` and `TestHashCacheDisabled`; this file wasn't exercising netsync or the production `SigCacheMaxSize` wiring.

node/blockchain/validate_test.go
jonathanMweiss: Done. Switched the assertions to `require.ErrorAs` / `require.Equal`, added `t.Parallel()`, and corrected the doc comment: `checkBlockScripts` skips via `IsCoinBase(tx)` on every transaction, not just index 0.

node/blockchain/validate_test.go
jonathanMweiss: Agreed — mining was only there to get past `checkProofOfWork` on the exported `CheckBlockSanity` path.

The test now calls `checkBlockSanity(..., BFNoPoWCheck)` with a vanity `0xdeadbeef` certificate, so it no longer depends on `-tags zkpow` or the FFI. Confirmed it passes without the zkpow tag.

node/txscript/hashcache.go
```diff
@@ -299,55 +298,46 @@ func NewTxSigHashes(tx *wire.MsgTx,
...
-// be present within the HashCache.
-func (h *HashCache) GetSigHashes(txid *chainhash.Hash) (*TxSigHashes, bool) {
-	h.RLock()
-	item, found := h.sigHashes[*txid]
-	h.RUnlock()
-
-	return item, found
+// ContainsHashes returns whether the partial sighashes for the passed
+// transaction currently exist within the HashCache.
+func (h *HashCache) ContainsHashes(txid *chainhash.Hash) bool {
```
Aharonee: `ContainsHashes` has no callers left anywhere in the tree — suggest removing it. This PR dropped the production one in `scriptval.go`, and `countCached` in the mempool spam test was the last remaining user until ecffc3f4 deleted that file.

Nothing will surface it otherwise: Go doesn't error on unused methods, staticcheck's `unused` skips exported identifiers by default, and there's no golangci config in the repo, so it would just sit there.

btcd parity isn't a reason to keep it either, given this PR already removes `AddSigHashes` and `GetSigHashes` — `HashCache`'s exported surface has already diverged to `LoadOrComputeSigHashes` / `ContainsHashes` / `PurgeSigHashes`. `PurgeSigHashes` earns its place (`scriptval.go` L270); this one no longer does.

One knock-on: removing it leaves `randEvictCache.contains` used only by the txscript tests, which is fine to leave as-is since those tests assert on cache internals.

Aharonee (dismissed):

PR #299, opened 2026-08-27
fix(zkpow): reject usize->u32/u16 narrowing of proof dimensions (CWE-681)

Aharonee:
## Summary

Fixes unchecked `usize` → `u32`/`u16` narrowing in `PlainProof::parse_proof` that bypasses the V3 Merkle leaf-count binding from #281 (CWE-681).

`check_declared_tree_sizes()` validates the wide (`usize`) declared dimensions, but `parse_proof` later narrowed them with unchecked `as` casts. On 64-bit targets a miner could add a multiple of `2^32` to `m` (or `n`/`k`), satisfy the leaf-count check on the inflated value, and have it silently wrap back to the intended public `u32` — reopening the carousel/work-amplification class of attack that #281 closed.

## Type

fix

## Scope

zk-pow

## Changes

- Reject `m`/`n`/`k` larger than `u32` and `noise_rank` larger than `u16` via checked `TryInto` **before** the tree-size check.
- Pre-validate MoE `e`/`top_k` fit `u16` before the tree-size check (which consumes the wide values through `total_b_cols` and routing leaf count).
- Use the checked values in `MiningConfiguration`/`PublicProofParams`.
- Replace unchecked `as u32` casts of A/B row indices and MoE inner A rows with checked conversions; use `checked_sub` for expert-offset inner B indices.
- Add regression test `zk-pow/tests/plain_proof_dimension_wraparound.rs`.

## Test plan

- [x] `cargo test --lib` (zk-pow): 86 passed, 2 ignored
- [x] `declared_tree_sizes` tests: 2 passed
- [x] Wraparound PoC test: 1 passed (previously returned `Ok`)
- [x] `cargo fmt --check` and `cargo clippy --lib --tests`: clean
- [ ] Existing tests pass (`task test`)
- [x] New tests added

## Credits

Reported by [@zcobi01](https://github.com/zcobi01). GHSA-vq6h-p843-m7c8.

### DIFF
--- a/zk-pow/src/ffi/plain_proof.rs
+++ b/zk-pow/src/ffi/plain_proof.rs
@@ -552,21 +552,13 @@ impl PlainProof {
         header: IncompleteBlockHeader,
         seed_derivation: SeedDerivation,
     ) -> Result<(PrivateProofParams, PublicProofParams)> {
-        // Reject every attacker-controlled `usize` dimension that does not fit its
-        // wire/public type *before* `check_declared_tree_sizes` validates the inflated
-        // values. On 64-bit targets, an unchecked `as` cast would otherwise truncate
-        // (wrap) an out-of-range dimension, letting a miner add a multiple of 2^width
-        // to a declared dimension, satisfy the leaf-count check on the wide value, and
-        // then have it silently wrap back to the intended public value (CWE-681).
+        // Reject oversized usize dims before tree-size checks so wraparound can't satisfy then truncate (CWE-681).
         let m: u32 = self.m.try_into().context("m exceeds u32")?;
         let n: u32 = self.n.try_into().context("n exceeds u32")?;
         let k: u32 = self.k.try_into().context("k exceeds u32")?;
         let noise_rank: u16 = self.noise_rank.try_into().context("noise_rank exceeds u16")?;
 
-        // Reject out-of-range MoE dimensions before the tree-size check below,
-        // which operates on the wide `usize` values (via `total_b_cols` and the
-        // routing leaf count). Otherwise an inflated `e`/`top_k` could satisfy the
-        // leaf-count check on the wide value and then wrap at the `u16` narrowing.
+        // MoE e/top_k also feed the tree-size check; reject overflow before wrap can hide it.
         let moe_config = match &self.moe {
             Some(mp) => Some(MoEConfig {
                 e: mp.e.try_into().context("e exceeds u16")?,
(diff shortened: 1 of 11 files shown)


o-k-d (dismissed):

zk-pow/src/ffi/plain_proof.rs
```diff
@@ -529,14 +552,35 @@ impl PlainProof {
         header: IncompleteBlockHeader,
         seed_derivation: SeedDerivation,
     ) -> Result<(PrivateProofParams, PublicProofParams)> {
-        let (m, n, k) = (self.m, self.n, self.k);
+        // Reject every attacker-controlled `usize` dimension that does not fit its
+        // wire/public type *before* `check_declared_tree_sizes` validates the inflated
+        // values. On 64-bit targets, an unchecked `as` cast would otherwise truncate
+        // (wrap) an out-of-range dimension, letting a miner add a multiple of 2^width
+        // to a declared dimension, satisfy the leaf-count check on the wide value, and
+        // then have it silently wrap back to the intended public value (CWE-681).
```
matan-dup: you have a single line of budget for this comment. trim.

zk-pow/src/ffi/plain_proof.rs
```diff
@@ -529,14 +552,35 @@ impl PlainProof {
...
+        let m: u32 = self.m.try_into().context("m exceeds u32")?;
+        let n: u32 = self.n.try_into().context("n exceeds u32")?;
+        let k: u32 = self.k.try_into().context("k exceeds u32")?;
+        let noise_rank: u16 = self.noise_rank.try_into().context("noise_rank exceeds u16")?;
+
+        // Reject out-of-range MoE dimensions before the tree-size check below,
+        // which operates on the wide `usize` values (via `total_b_cols` and the
+        // routing leaf count). Otherwise an inflated `e`/`top_k` could satisfy the
+        // leaf-count check on the wide value and then wrap at the `u16` narrowing.
```
matan-dup: trim comment. Tell me why, not what or how.

matan-dup (dismissed):

matan-dup (dismissed):

zk-pow/src/ffi/plain_proof.rs
Aharonee: Rewrote to a single why-line: leaf-count binds `usize`, public dims are `u32`/`u16`, wrap would unbind them.

zk-pow/src/ffi/plain_proof.rs
Aharonee: Dropped the how/what MoE comment; the one-line invariant above covers `e`/`top_k` too.

matan-dup (approved):

PR #308, opened 2026-09-02
fix(netsync): verify header connectivity before ZK certificate

jonathanMweiss:
## Summary

Reorder validation so less expensive checks reject bad headers and blocks *before* the
compute intensive ZK certificate verification runs.

## Type

fix

## Scope

node — `netsync` (`handleHeadersMsg`), `blockchain` (`checkBlockSanity`)

## Why it matters
slightly hardening the process of cert verification.

## Test plan

- [x] Existing tests pass (`task test`) — `go test -tags=xmss,zkpow,rpctest
  ./node/blockchain/... ./node/netsync/...` green with this commit standing
  alone; `go build ./...` clean.
- [ ] New tests added (if applicable)

### DIFF
--- a/node/blockchain/validate.go
+++ b/node/blockchain/validate.go
@@ -419,11 +419,6 @@ func checkBlockSanity(block *btcutil.Block, chainParams *chaincfg.Params, timeSo
 
 	flags |= NetBehaviorFlags(chainParams)
 
-	err := CheckBlockHeaderSanity(header, cert, chainParams.PowLimit, timeSource, chainParams.MaxTimeOffsetMinutes, flags)
-	if err != nil {
-		return err
-	}
-
 	// A block must have at least one transaction.
 	numTx := len(msgBlock.Transactions)
 	if numTx == 0 {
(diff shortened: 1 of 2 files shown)


o-k-d (commented):
It is debatable if it is the correct order.
All other checks are running without the peer proving any pow from his side.
Making a certificate prove pow costs something to the prover -- and if veritication fails, a ban is triggered.

jonathanMweiss:
Both a bogus proof and creating a malformed header are relatively easy to fabricate; benchmarking the costs:
- Structural checks without ZK verification costs ~550 ns. 
- Inspecting a malformed ZK proof costs 3.7~4ms (I took a valid ZK proof and flipped one of its bits).

Either way the attacker is disconnected from the sync-peer position; which, as an inbound peer, is relatively costly to regain. 
I don’t consider either ordering strictly “correct;” the “costly checks last” guidance from several advisories was the main pressure to reorder, since cheap-first reduces the DoS surface (garbage never reaches the ZK). 
Or and I considered a ban score but decided against it for now, since losing the sync-peer slot is already the penalty.

PR #317, opened 2026-09-10
fix(wallet): SPV sends no longer get stuck pending; users can rebroadcast or remove pending transactions

Aharonee:
## Summary

Fixes a bug in the oyster wallet's SPV publish path (`--usespv`): a send published while no peer requested the transaction was reported as successful and recorded as an unconfirmed spend, although nothing had left the machine. The transaction could never appear in a mempool or block, its inputs stayed locked, and the only recovery was `dropwtxmgr` or wipe-and-restore. This affects any oyster running in SPV mode, whether driven over RPC, from `oystercli`, or embedded in the Desktop Wallet, where it was reported.

The fix has two parts:

1. **No verdict, no record.** The SPV layer reports such a broadcast as `NotRelayed`, oyster fails the send cleanly and keeps the UTXOs spendable, and `getsyncprogress` exposes the peer count so clients can refuse to send while peerless.
2. **Pending transactions are the user's to manage (Electrum's model).** Under SPV oyster announces a transaction exactly once, when it is sent or when the user asks, and never re-announces on its own: an SPV node cannot observe mempools, so a background retry loop only hides whether the network ever held the transaction from the person who has to decide what to do about it. Instead, every unconfirmed transaction shows whether a peer took it after an announcement made in this session ("relayed 5m ago" / "not announced since start"), and two explicit actions exist on every surface: `rebroadcasttransaction` (announces pending ancestors first) and `removetransaction` (drops the transaction and its pending dependents, freeing the inputs, behind a double-spend warning). The Neutrino per-block/per-minute rebroadcast loop is removed; the full-node backend is unchanged, since there the node's mempool plays the Electrum server's role.

## Type

fix

## Scope

wallet (oyster), spv, node/btcjson, oystercli, desktop wallet

## Root cause

`spv/query.go` `sendTransaction` returned `nil` when `len(replies) == 0`, i.e. when no peer sent `getdata` for the announced txid (instantly with zero connected peers, after the 5s broadcast timeout with connected-but-silent peers). `reliablyPublishTransaction` persists the spend before broadcasting and `publishTransaction` treats a bare `nil` as success, so the record stayed. Rebroadcast did not heal it: the retry set was in-memory and lost on restart, and `resendUnminedTxs` went through the same silent path. Clients also had no way to see the peer count: `getinfo` is unavailable under SPV and `getsyncprogress` reported heights only.

### Fail unrelayed sends

- `37ec3090` **fix(spv)**: `sendTransaction` returns `pushtx.NotRelayed` when there are no connected peers or none sent `getdata`; the reason distinguishes the two. Transactions are announced with `MSG_TX`, as BIP-144 reserves the witness inventory types for `getdata`; pearld already accepted the witness type, so this is hygiene rather than the incident cause.
- `0ca3c967` **fix(wallet)**: `chain.ErrTxNotRelayed`, mapped from the SPV error with its reason preserved. A first publish drops the record and returns the error (safe: an inv carries only the txid, so with no `getdata` nothing that could confirm left the node); a resend keeps it.
- `d6a9dfa7` **feat(wallet)**: `getsyncprogress` gains `connections`, oyster's connected peer count.
- `a2dcd977` **fix(desktop)**: the send page maps the not-relayed error (no peers, or none that asked) to "nothing was sent, funds untouched" and words success as "broadcast to the network". Its pre-flight peer poll was dropped again in `61f37489`: the daemon already fails a peerless send instantly and safely, so the poll only moved the same message before the click.
- `efee102a` **refactor(wallet)**: `publishMode` (`publishNew`, `republish`) replaces a bool; the unrelayed case gets its own log line; `removeUnminedTx` is shared by the removal sites.
- `96d0475b` **test**: prune low-value tests.
- `1912982c` **chore**: desktop wallet `2.0.5 → 2.0.6`; `dc5ad315` bumps pearl to `1.4.9`, since master tagged `v1.4.8` with #324 while this PR was open.

### Explicit pending transaction controls

- `a55aada4` **refactor(spv)**: `pushtx.Broadcaster` (goroutine re-announcing every pending tx per block and per minute) is replaced by a one-shot `pushtx.Publisher` that records, per session, when a peer last requested each transaction; `ChainService` exposes it as `LastRelayed`/`ForgetTransaction` and confirmed transactions drop out of the map from the rescan path. The rescan-triggered `resendUnminedTxs` now runs only for full-node backends. Deleting `broadcaster_test.go` also removes the flaky `TestRebroadcast`.
- `53f6fe66` **feat(wallet)**: `Wallet.RemoveTransaction` (pending only; takes every pending dependent, frees inputs, forgets relay evidence) and `Wallet.RebroadcastTransaction` (pending unmined ancestors first, in dependency order, so a child whose parent no peer holds cannot sit as an orphan). New optional `chain.BroadcastTracker` capability, implemented by `NeutrinoClient` only; `listtransactions`/`gettransaction` carry `relayed`/`lastrelaytime` for unconfirmed transactions on such a backend.
- `d1c2f185` **feat(rpc)**: `removetransaction` and `rebroadcasttransaction`; help text states that "no peer requested it" on a rebroadcast is ambiguous by construction (peers that already hold a transaction stay silent) and that removing a transaction a peer holds enables a double-spend attempt.
- `990c6504` **feat(oystercli)**: history rows and the detail view label pending transactions as "pending, relayed <ago>" or "pending, not announced since start"; the detail view offers Rebroadcast and Remove (the latter behind a confirm with the double-spend caveat); Node & sync reads the peer count from `getsyncprogress`, which works under SPV where `getinfo` fails.
- `d25ee569` **feat(desktop)**: Activity page shows the same status and offers Rebroadcast / Remove, the latter through a native confirm; the dashboard preview shows the status only.
- `f28e42d8` **refactor**: post-review trim (`RemoveTransaction` reports what wtxmgr actually deleted, one `decodeTxHash`, one `formatTimeAgo`, shared test fakes).
- `3147f82e` **fix**: relay status and the two actions apply to sends this wallet made. An incoming unconfirmed payment was never announced by this daemon, so `relayed: false` would read as "stuck" rather than "received, unmined"; it now shows plain pending and offers no rebroadcast/remove in oystercli or the Desktop (the RPCs still accept any unmined txid).
- `aaaac696` **fix(desktop)**: the no-peers notice no longer sticks after peers return (review finding; the check itself was removed in `61f37489`).
- `1127e164` **docs**: RPC help and oystercli README say "send"; the Activity notice wording follows the action it reports.
- `e4a60fec` **test**: drop the trivial `isWalletSend` unit test. `c3830191` merges `master` (#324); no conflicts.
- `61f37489` **refactor(desktop)**: drop the pre-flight peer poll and the above-the-list Activity notice; the desktop no longer reads `connections` at all.

## Test plan

- [x] Existing tests pass: `go test ./spv/... ./wallet/chain/... ./wallet/rpc/... ./wallet/wallet/... ./node/btcjson/... ./wallet/cmd/oystercli/...`, and `go test -race -tags=xmss,zkpow` on the touched packages. `spv/pushtx` now passes `-race -count=25` in full.
- [x] Regression tests: `TestGhostPendingSendRegression` (drives `SendOutputs` through a real peerless `neutrino.ChainService`; reproduces the incident on the unfixed base), `TestSendTransactionNoPeers`, `TestTransactionInvAnnouncesByTxid`, `TestNeutrinoClientSendRawTransactionNotRelayed`, `TestPublishTransactionNotRelayed`.
- [x] New tests: `pushtx` `TestPublisher*` (success and `Mempool` record evidence, `NotRelayed` does not, `Forget`, concurrent use under `-race`); `wallet` `TestRemoveTransaction` (drops dependents, frees inputs, refuses mined, works without a tracker), `TestRebroadcastTransaction` (parent before child, `ErrTxNotRelayed` keeps the record, a rejection removes it), `TestRelayStatusInListings`, `TestUnminedAncestry`, `TestResendAfterRescanBackendGate`; `legacyrpc` `TestPendingTxError`; `oystercli` `TestRelayLabel`, `TestFmtAgo`, `TestFmtPeerCount`, `TestIsNotRelayedError`, `TestPendingTxCalls`.
- [x] Every new test run 25x under `-race` with no failure.
- [x] Manual pass on simnet (oyster SPV + pearld, over RPC and through the `oystercli` TUI): fresh send shows `relayed: true` with a timestamp; after a daemon restart the same transaction reads `relayed: false` and nothing is resent after the rescan; `removetransaction` frees the input (UTXO count 4→5) and the record is gone; `rebroadcasttransaction` against a peer with an empty mempool returns `announced: [txid]` and the tx appears in pearld's mempool, and against a peer that already holds it returns the ambiguous not-relayed error with the record kept; confirmed txid → `-8`, bad hex → `-22`, unknown → `-5`. Also observed the caveat the Remove warning describes: a second spend of an input freed while pearld still held the first tx was rejected with `txn mempool conflict` and dropped by the wallet.
- [x] Desktop: `tsc --noEmit` adds no errors over the pre-existing baseline.
- [x] `go generate` for RPC help; `TestRPCMethodHelpGeneration` green.

## Notes

- Behaviour changes at oyster's RPC: `sendmany`/`sendtoaddress`/`sendfrom` return an error (`transaction not relayed to any peer: ...`) instead of a txid when no peer requested the transaction; `listtransactions`/`gettransaction` gain `relayed`/`lastrelaytime` for unconfirmed sends under SPV (omitted for incoming payments and on a full-node backend); new `removetransaction` and `rebroadcasttransaction`. Sends that reach at least one peer are unchanged.
- Under SPV nothing re-announces after a restart any more: a pending transaction from a previous session reads "not announced since start" until the user rebroadcasts it or it confirms. Relay evidence is per session and not persisted, matching how Electrum derives its `Local` state live.
- "Success" means "at least one peer requested the transaction", not "in a mempool"; with `DisableRelayTx` that is the strongest signal an SPV client has. `relayed: false` therefore means "not announced since start", not "the network lacks it".
- Follow-ups (out of scope): R3 detection (a peer that dropped while the tx sat in its send queue still counts as a relay; `peer.outHandler` signals `doneChan` on both success and failure, so this needs a `chan error`), a bounded wait-for-peers before failing a first publish, RBF/`bumpfee` as a separate policy decision. Found during the manual pass, pre-existing: #321 (oystercli history order), #322 (stale `BlockStamp` after restart).


jonathanMweiss:
The fix is worth landing: a send that no peer requested now fails and
releases its inputs instead of reporting success.

Three gaps on the paths it touches:

- **`republish` keeps a record for a rebroadcast that may never come.** The
  Broadcaster only admits transactions whose broadcast succeeded, and
  `resendUnminedTxs` only fires on rescan-finished — so after a restart a kept
  transaction sits in neither retry loop while still locking its inputs.
  (Inline comment filed.)

- **The no-peers error never clears.** Once set it survives reconnection, and
  `!!error` keeps the send button disabled until the user edits a field.
  (Inline comment filed.)

- **Success is still weaker than the new banner implies.** `replies` is recorded
  when a peer sends getdata and we enqueue the tx, but `doneChan` can't
  distinguish a successful socket write from a failed one
  (`node/peer/peer.go:1800-1838`), so a peer that dropped while the message sat
  queued still counts as a relay. Changing it to `chan error` is a small fix.

Out of scope here, worth its own issue: for a transaction a peer *did* take but
that never confirms, there's still no terminal failed state; the inputs stay
locked indefinitely, with no cancel and no fee bump.

## Suggestions

- **Support Replace-By-Fee (RBF; BIP-125).** Lets a client replace a stuck
  transaction with a higher-fee version spending the same inputs. pearld's
  mempool already accepts replacements by default; the wallet just never opts in,
  leaving `Sequence` at `wire.NewTxIn`'s `0xffffffff`. The code change is one
  line, but it's a policy call, not a free win: every Pearl transaction becomes
  replaceable, which weakens assurance for anyone accepting zero-conf payments.
  The real work is the `bumpfee` RPC and the UI on top. Bitcoin Core
  (`bumpfee`), Electrum, Sparrow, and BlueWallet all offer this, and consumer
  wallets that hide the UTXO model generally rely on it to resolve exactly this
  failure mode rather than exposing the mechanics.
  
 - **`resendUnminedTxs` runs once, and never again until a restart.** The mechanism should run inside a single goroutine in select loop that is triggered via channel on a new block notification/ ticker.  it should listen on the wallet's quit channel for closure. (this is for "republish` keeps a record for a rebroadcast that may never come.").

## maybe

maybe we can confirm delivery by challenge? Pick one peer at random and exclude
  it from the broadcast, then ask *that* peer for our own txid with a `getdata`.
  If it returns the transaction, the tx reached at least one node through the network
  rather than from us — some evidence it started propagating. 
  If it returns `notfound`,  it hasn't arrived yet; retry over a finite broadcast window.

Aharonee:
Thanks, agreed on all three gaps. The PR has grown since your review (commits from `a55aada4` on), which changes the answer to the first one. Point by point:

**`republish` keeps a record for a rebroadcast that may never come.** Resolved, but not with another loop. Under SPV oyster now announces exactly once (on send, or when the user asks) and never re-announces on its own: the Neutrino `Broadcaster` is gone (`a55aada4`) and the post-rescan resend runs only for the full-node backend. Every pending tx shows whether a peer took it after an announcement in this session ("relayed 5m ago" / "not announced since start"), and the two ways out are explicit: `rebroadcasttransaction` (pending ancestors first) and `removetransaction` (drops the tx and its pending dependents, frees the inputs, behind a double-spend warning) — `53f6fe66`, `d1c2f185`, exposed in RPC, oystercli and the Desktop. So a kept record no longer sits in "neither loop"; there is no loop, the user sees the state and decides.

**No-peers error never clears.** Right. Fixed in `aaaac696`: the submit-time check no longer writes into the form error; the notice and the disabled button follow the polled peer count and clear when a peer returns.

**Success weaker than the banner.** Agreed; it's R3 in the RCA. `peer.outHandler` signals `doneChan` in both the success and the failure branch, so waiting on it wouldn't help — it needs a `chan error`, which touches `node/peer`. I'd do it as its own PR rather than fold it in here.

**Terminal failed state / cancel / fee bump.** Cancel and re-announce are the two actions above. Fee bump: see RBF.

**RBF.** Agree it's a policy decision, not a free line: signalling makes every Pearl tx replaceable, and the real work is `bumpfee` + UI. Not in this PR; worth its own issue with that trade-off spelled out. `removetransaction` already gives the manual half (respend the freed inputs), and the simnet run shows its limit: a respend while pearld still held the original was rejected as `txn mempool conflict`, exactly the case RBF would resolve.

**`resendUnminedTxs` in a select loop.** This is where I went the other way, deliberately. An SPV node can't see mempools, so a wallet-side retry loop only hides from the user whether the network ever held the transaction. Electrum announces once and makes re-announcing a user action; that's the model adopted here. The full-node backend keeps the resend, since there the node's mempool plays the Electrum server's role. If you think a loop is still needed on top of that, let's discuss.

**Delivery challenge (withhold from one peer, `getdata` from it).** Nice idea, and complementary to the per-session relay evidence: it would give an actual propagation signal instead of "a peer asked". Caveats: inv trickling means the window has to be several seconds and `notfound` isn't a verdict, so it's a bounded probe rather than a check. Follow-up candidate.

One more thing: the two inline comments you mention aren't visible on the PR, probably still in a pending review — could you submit it?

spv/query.go
```diff
@@ -957,20 +984,21 @@ func (s *ChainService) GetBlock(blockHash chainhash.Hash,
...
-		invType = wire.InvTypeTx
+
+	numPeers := len(s.Peers())
+	if numPeers == 0 {
+		return notRelayedError(tx.TxHash(), numPeers)
 	}
 
-	// Create an inv.
+	// Announce the transaction by txid. A peer can request its preferred
+	// serialization in getdata, which we answer using qo.encoding below.
```
jonathanMweiss: there's a lot of added comments which aren't very helpful in this file.

wallet/cmd/oystercli/transactions.go
```diff
@@ -125,6 +130,118 @@ func showTransactionDetail(c *client, txid string) error {
...
+	// confirm on chain.
+	if tx.Confirmations > 0 || !isWalletSend(tx) {
+		return nil
+	}
+	return pendingTxActions(c, tx.TxID)
+}
+
+// isWalletSend reports whether this wallet originated the transaction, as
+// opposed to merely receiving an output from it.
+func isWalletSend(tx *btcjson.GetTransactionResult) bool {
```
jonathanMweiss: isWalletSend and its doc comment ("whether this wallet originated the transaction") claim more than the check does. A "send" detail only means the transaction spends an output this wallet owns (len(details.Debits) > 0 in gettransaction). It doesn't mean this daemon created or broadcast it. Against a pearld backend, mempool transactions from another device using the same seed also show up as pending sends. Under SPV, a reorg can do the same.

Could we rename it to say what it checks, e.g. spendsWalletCoins?

jonathanMweiss (dismissed):
I reviewed the SPV and wallet changes; seems good. 

two nits.

wallet/cmd/oystercli/transactions.go
Aharonee: Done in c0b31d82 — renamed to `spendsWalletCoins` and tightened the doc to “spends wallet-owned outputs” (not “this daemon originated/broadcast it”). Predicate unchanged.

spv/query.go
Aharonee: Done in 081e3724 — trimmed the options preamble and the “Announce the transaction by txid…” block in `spv/query.go`. Left one-line godoc on `newTransactionInv` / `notRelayedError`; getdata/reject callback comments unchanged.

Aharonee:
@jonathanMweiss thanks for the review. Your approval was dismissed when the nit fixes were pushed, so could you take another look? Two changes since then:

- `761ddbda` puts back three comments that `081e3724` had dropped from `spv/query.go`: the `newTransactionInv` doc, the options preamble, and the announce-by-txid note. They are upstream neutrino's wording for code identical to upstream ([lightninglabs/neutrino#386](https://github.com/lightninglabs/neutrino/pull/386)), kept verbatim so future syncs merge cleanly. What this PR adds of its own in that file is down to the two-line note on `notRelayedError` and the `NotRelayed` clause in `sendTransaction`'s doc. The `spendsWalletCoins` rename is in as you suggested.
- `3b21ebc0`: a rejected rebroadcast deletes the record, so its inline error had no row left to render on and the transaction just vanished from Activity. Rebroadcast/Remove errors now open a dialog.

jonathanMweiss:
reviewed the go code. comments changed. LGTM.

Aharonee:
@jonathanMweiss thanks for the LGTM. It went in as a comment, so the PR still needs an approving review from you.

One correction first, on your point that the no-peers error never clears. Once the pre-flight peer check was dropped (`61f37489`), a send with no peers surfaces as the daemon's not-relayed error, and that error still disabled Send until a field was edited, so the stuck state you described was back. `64237bdd` keeps Send enabled after it. I checked on a local simnet that clicking Send again once peers return goes through without touching the form.

`f571fc88` also covers the latest Bugbot note: the rebroadcast error dialog now says the funds are spendable again only when the reload shows the record is gone, instead of guessing from the error text.

wallet/wallet/wallet.go
```diff
@@ -3837,17 +3877,8 @@ func (w *Wallet) publishTransaction(tx *wire.MsgTx) (*chainhash.Hash, error) {
...
-		txRec, err := wtxmgr.NewTxRecordFromMsgTx(tx, time.Now())
-		if err != nil {
-			return err
-		}
-		return w.TxStore.RemoveUnminedTx(txmgrNs, txRec)
-	})
-	if dbErr != nil {
-		log.Warnf("Unable to remove invalid transaction %v: %v",
-			tx.TxHash(), dbErr)
+	if err := w.removeUnminedTx(tx); err != nil {
```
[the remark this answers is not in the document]
Aharonee: Fixed in `ec2fda1c`: an explicit rebroadcast now keeps the transaction when a peer rejects it and reports the error, as Electrum and Bitcoin Core do. Dropping it is left to Remove, which carries the double-spend warning. `1a2fa42e` removes the Desktop copy that said the funds were free. The automatic full-node resend still removes rejected records, since its verdict comes from the wallet's own node.

Aharonee:
@jonathanMweiss one Go change since your LGTM: `ec2fda1c` stops an explicit rebroadcast from deleting a transaction a peer rejects (the security review above). Electrum and Bitcoin Core do the same: they report the rejection and keep the transaction, and removing it is the user's call. The automatic full-node resend is unchanged. Could you take a look and approve?

PR #324, opened 2026-09-16
feat(node): absorb upstream btcd hardening and robustness changes

Aharonee:
## Summary

Ports the applicable changes from btcsuite/btcd v0.26.0–v0.26.2 into `node/`, adapted by hand to Pearl's certificate PoW and mandatory v2 transport (no history merge). Also adds the mainnet checkpoint at height 110000 and bumps the version to 1.4.8.

## Type

feat

## Scope

node

## Changes

PR numbers below refer to btcsuite/btcd.

- **Wire / P2P** (PRs 2474, 2479, 2491, 2504, 2515, 2518, 2536) — strict message decoding (trailing bytes rejected), `MaxProtocolMessageLength` enforced on every read and write path, addrv2 handling of IPv4-mapped IPv6, additional unroutable ranges in `addrmgr`, panic recovery in peer goroutines, tighter connection association and peer lifecycle handling, inv handling while no sync peer is selected, with low-quality peers routed through the existing getheaders probe. The protocol cap is 8,000,000 bytes rather than Core's 4 MB because Pearl headers carry certificates; `MsgBlock`'s cap includes `CertificateMaxSize`.
- **Inbound admission** (PR 2576) — new `internal/inbound` package: bounded inbound connection count, per-source pending-handshake budget, leased v2 responder crypto phases; `connmgr.MaxInbound`; `maxpeers` validated at config load; deterministic outbound targeting. v1-fallback paths skipped since Pearl has no v1 transport.
- **Decoding** (PRs 2558, 2575) — `NewBlockFromBytes`/`NewTxFromBytes` and RPC decode paths reject trailing data; own-database blocks load via the lenient `DBBlockFromBytes`; strict hash parsing in `chainhash`.
- **Crypto & PSBT** (PRs 2539, 2545, 2561, 2563, 2567, 2601) — canonical range checks in `schnorr.ParseSignature` and `DecodeWIF`; musig2 decode error propagation; PSBT nil guards, strict base64, trailing-data checks, `multi_a` finalizer.
- **RPC client** (PR 2500) — context-aware POST cancellation, batch handling, `DisableAuth`, cached endpoint URL; a failed batch `Send` now fails only the requests it submitted, so a concurrent batch's requests are neither failed with the wrong error nor orphaned.
- **Storage** (PR 2425) — treap node pool and `PutPairs` bulk insert for the ffldb cache; indexer `dropIndex` error propagation.
- **Database contract** — `Tx.FetchBlock*` bytes are documented as caller-owned and valid after the transaction ends (what ffldb has always provided and every caller relies on); a backend on reused or mapped memory must copy at its own boundary. No copies at call sites.
- **Misc** — regtest sync candidates no longer localhost-only (PRs 2487, 2548); rpctest build race fix; `txscript` uses `text/template`.
- **Checkpoint** — mainnet checkpoint at height 110000 (`c119b81d…d50630`), verified against Blockbook.
- **Version** — 1.4.7 → 1.4.8.

## Test plan

- [x] Existing tests pass (`task test:go`, full suite with `-race -tags=rpctest,xmss,zkpow`)
- [x] New tests added (wire, peer, inbound, netsync, indexers, treap, rpcclient, rpcserver, chainhash, psbt)


jonathanMweiss:
DBBlockFromBytes is untested.

jonathanMweiss:
suggestion; set the following code from `CompleteHandshake` in helper function, so release can be called via a `defer()`.

```Go
release := func() {}
	if !initiating {
		release, err = p.acquireResponderAdmission()
		if err != nil {
			return err
		}
	}

	// Calculate the shared secret to be used in creating the packet
	// ciphers.
	ecdhSecret, err := ellswift.V2Ecdh(
		p.privkeyOurs, ellswiftTheirs, p.ellswiftOurs, initiating,
	)
	if err != nil {
		release()
		log.Errorf("Failed to calculate ECDH shared secret: %v", err)
		return err
	}

	log.Tracef("Calculated ECDH shared secret: %x", ecdhSecret)

	err = p.createV2Ciphers(ecdhSecret[:], initiating, prlnet)
	release()
	if err != nil {
		return err
	}
```



Right now, a panic before a release can be recovered and lock not release a global resource limiter. 

Suggestion:
```Go

// deriveV2Ciphers holds the responder admission lease for exactly the CPU-bound ECDH and cipher setup; the
// deferred release cannot leak the lease if either panics.
func (p *Peer) deriveV2Ciphers(ellswiftTheirs [64]byte, initiating bool, prlnet PearlNet) error {
	release := func() {}
	if !initiating {
		var err error
		release, err = p.acquireResponderAdmission()
		if err != nil {
			return err
		}
	}
	defer release()

	// Calculate the shared secret to be used in creating the packet
	// ciphers.
	ecdhSecret, err := ellswift.V2Ecdh(
		p.privkeyOurs, ellswiftTheirs, p.ellswiftOurs, initiating,
	)
	if err != nil {
		log.Errorf("Failed to calculate ECDH shared secret: %v", err)
		return err
	}

	log.Tracef("Calculated ECDH shared secret: %x", ecdhSecret)

	return p.createV2Ciphers(ecdhSecret[:], initiating, prlnet)
}```

jonathanMweiss:
The handleGetBlock fix (slices.Clone at rpcserver.go:1066) conforms with the database.Tx contract, but four other spots in this package still let transaction-bound bytes escape the DB.View closure:

- node/rpcserver.go:2756 (handleGetRawTransaction) — txBytes from dbTx.FetchBlockRegion(blockRegion) is used after the View ends: hex-encoded at :2767 and deserialized at :2780.
- node/rpcserver.go:3066 (fetchInputTxos) — txBytes from dbTx.FetchBlockRegion(blockRegion) is deserialized at :3075, after the View closure returns.
- node/rpcserver.go:3361 (handleSearchRawTransactions) — serializedTxns from dbTx.FetchBlockRegions(regions) are stored in retrievedTx{txBytes} and appended to the outer addressTxns slice; they're hex-encoded at :3419 and deserialized in the verbose path, all after the View ends.
- node/server.go:1722 (pushBlockMsg) — blockBytes from dbTx.FetchBlock(hash) is deserialized at :1737 after the View closure returns.

Per node/database/interface.go:262, data returned by FetchBlock/FetchBlockRegion(s) is only valid during the transaction (undefined behavior after it ends). Fix: same pattern as handleGetBlock — slices.Clone inside the closure (per-slice for FetchBlockRegions). Latent with today's ffldb (fresh allocation on read), but ffldb does alias pending-block data for FetchBlockRegion on a writable tx (ffldb/db.go:1342-1353), so contract conformance matters.

jonathanMweiss:
Actually, we should either remove the panic catching inside the peer's goroutines or ensure that on panic they'll call their cleanup + disconnect() methods to ensure a clean exit.

Aharonee:
Addressed in `3bff7e4b` (on top of `54681536`/`ceb344da`):

- **Witness invs bypass the gate (Bugbot):** the `lastBlock` scan and the low-quality skip now treat `InvTypeWitnessBlock` like `InvTypeBlock`; covered by a new case in `TestInvGateWithoutSyncPeer`.
- **`DBBlockFromBytes` untested:** kept the coverage from `54681536` but as a table-driven testify test (exact / trailing bytes dropped / truncated / empty, plus the strict constructor still rejecting the trailing-bytes input), matching the rest of the tests added in this PR.
- **`defer release()` in `CompleteHandshake`:** `ceb344da` landed the `deriveV2Ciphers` extraction; added `TestDeriveV2CiphersReleasesLease` pinning release on success, release when the derivation panics, and no lease for the initiator.
- **Transaction-bound bytes escaping `View`:** `slices.Clone` inside the closure at `handleGetRawTransaction`, `fetchInputTxos`, and `handleSearchRawTransactions` (per region). `pushBlockMsg` decodes the block inside the view instead of copying — the decoded `MsgBlock` owns its memory, so the serve path pays no block-sized copy. `TestPushBlockMsgDecodesInsideView` uses the invalidating fake DB, which now poisons bytes with `0xff` after the view: all-zero bytes still decode as an empty block, so zeroing could not catch a late decode.

Unchanged from the earlier discussion: the `blockchain` package (`dbFetchBlockByNode` and its callers) still carries DB bytes out of views on hot paths; the contract decision there stays a follow-up.

Aharonee:
Re the recover placement — went with the first option in `9f42edbf`, back to upstream's two sites plus one fix:

- Removed the recover from `stallHandler`, `queueHandler`, `outHandler`, `pingHandler`. Nothing untrusted reaches them, so a panic there is a broken invariant; recovering skipped their non-deferred drain loops, leaving `doneChan` waiters and `stallControl` senders hung while the node kept running. Better to crash. The deferred `close(queueQuit)`/`close(outQuit)` from the earlier round stay.
- `inHandler` keeps its recover: it is where wire decoding and the listener callbacks run, and its exit path (`close(inQuit)`, `Disconnect`) is entirely deferred, so the exit is clean.
- The negotiation goroutine now reports a recovered panic through `negotiateErr`, so `start()` disconnects immediately instead of waiting out the 30 s `negotiateTimeout` with the slot held. `TestStartReportsNegotiationPanic` pins it (fails against the previous code at the 5 s bound); `TestQueueHandlerRecoversFromPanic` is gone with the behavior it pinned.

Aharonee:
Update on the `FetchBlock*` byte lifetimes — `f425521d` resolves this at the contract instead of at the call sites, reversing what landed in `3bff7e4b`:

- `database.Tx` now documents `FetchBlock`/`FetchBlocks`/`FetchBlockRegion(s)` bytes as caller-owned and valid after the transaction ends; a backend reading from reused or mapped memory must copy at its own boundary. The same-transaction case is stated explicitly: a block stored earlier in a writable tx comes back as the bytes passed to `StoreBlock`.
- The four `slices.Clone`s (`getblock` included) are removed and `pushBlockMsg` is back to master's shape; those sites now diff clean against `master`. Tests that modelled a buffer dying with the transaction are gone; `getblock` keeps a test that own-DB trailing bytes are stripped from the response.

Why: ffldb has allocated on every read since the package was written, and every caller in `blockchain`, `netsync`, and the RPC server has always carried the bytes out of the view — the borrowed contract was honored nowhere and enforced by nothing. Honoring it would mean a second copy at each escaping site, including `dbFetchBlockByNode` on reorg and relay paths, to keep room for a memory-mapped backend that never existed. If zero-copy reads are ever wanted, the fit is an additive scoped accessor (`ViewBlock(hash, func([]byte) error)`) next to the owned default, not a borrowed `[]byte` return.

Aharonee:
One more commit outside the review threads: `f9c70de7` fixes a batch-mode bug in the ported rpcclient code (present upstream too). `failBatchRequests` resolved every request in `batchList` with the failed `Send`'s error and then reset the request map and lists; a request queued for a concurrent batch shares `batchList`, so it received the other batch's error and was orphaned — its own response then matched nothing, the result was dropped, and that `Send` still returned `nil`. `sendAsync` now returns the ids it marshalled and the failure path resolves exactly those via `removeRequest`, so no shared state is reset. `TestBatchSendFailureSparesConcurrentBatch` reproduces the scenario and fails against the previous code.

PR description updated for this and for the `Tx.FetchBlock*` contract change.

